PR #158, opened 2018-04-30
[WIP] Create top-level root views to separate logged in and out

sindresorhus:
This adds a `RootContainer` which manages the whole app. This changes the responsibility of `AppContainer` to only handle the logged-in app. I thought this would be a cleaner separation, but I'm not entirely sure this is the correct approach. I'll need to think about it more. The task of the `RootContainer` is to change between the `Login` and `App` views depending on whether the user is logged-in or not.


sindresorhus:
I don't think this is worth the added complexity. Closing.

PR #192, opened 2018-05-09
Improve sell order handling

### LINKED ISSUE #184: Improve sell order handling
`marketmaker` doesn't have any concept of sell orders due to the any/any order book combinations.

If you are on the `CHIPS/KMD` pair and place a sell order, `marketmaker` will place a `KMD/CHIPS` buy order.

This works but it's not very clear in the UI. We need to clearly show buy and sell orders.

One potential solution here: https://github.com/lukechilds/hyperdex-bugtracker/issues/15

lukechilds:
Fixes #184 and fixes lukechilds/hyperdex-bugtracker#15

See https://github.com/lukechilds/hyperdex-bugtracker/issues/15#issuecomment-387599771 for info.

Also, this breaks the `<SwapItem/>` component on the Trades view. I tried copying the new markup over but the it gets the colours styles but remove the positioning styles. I looked in the Trades styles but couldn't see anything regarding to positioning. Are you able to have a look at this.

It also seems confusing that the same component is used in two views and borrows some styles from both the views but then also has some unique styles. I think we should definitely consolidate this into a single component used in both places. It should display the same everywhere with both the cancel button and the view button in both places.

### DIFF
--- a/app/renderer/views/Exchange/Swaps.js
+++ b/app/renderer/views/Exchange/Swaps.js
@@ -72,10 +72,14 @@ const All = () => (
 const Split = () => {
 	const {state} = exchangeContainer;
 
-	const filteredData = state.swapHistory.filter(x =>
-		x.baseCurrency === state.baseCurrency &&
-		x.quoteCurrency === state.quoteCurrency
-	);
+	const filteredData = state.swapHistory.filter(swap => {
+		const tradingPair = [state.baseCurrency, state.quoteCurrency];
+
+		return (
+			tradingPair.includes(swap.baseCurrency) &&
+			tradingPair.includes(swap.quoteCurrency)
+		);
+	});
 
 	return (
 		<SwapList swaps={filteredData}/>
--- a/app/renderer/views/Exchange/Swaps.js
+++ b/app/renderer/views/Exchange/Swaps.js
@@ -38,8 +38,8 @@ const SwapItem = ({swap}) => (
 	<div className="row">
 		<div className="timestamp">{formatDate(swap.timeStarted, 'HH:mm DD.MM')}</div>
 		<div className="pairs">{swap.baseCurrency}/{swap.quoteCurrency}</div>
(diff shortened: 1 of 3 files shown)


sindresorhus:
> Also, this breaks the <SwapItem/> component on the Trades view. I tried copying the new markup over but the it gets the colours styles but remove the positioning styles. I looked in the Trades styles but couldn't see anything regarding to positioning. Are you able to have a look at this.

Should be fixed by https://github.com/lukechilds/hyperdex/pull/192/commits/5f365e80fe385dadc7912d16a80f20bd698356d7

lukechilds:
Thanks, 5f365e8 fixed it. Pretty sure that's exactly what I did. 🤷‍♂️

>There are multiple reasons for this

Yeah I remember the discussions, not blaming you 😉Just saying now would be a good time to consolidate so we don't duplicate more work in each view.

>I'd be happy to change it now, but it's going to need a lot of conditionals.

Awesome, why will it need a lot of conditionals? Can't they just both use the same CSS grid styles?

sindresorhus:
> Yeah I remember the discussions, not blaming you 😉Just saying now would be a good time to consolidate so we don't duplicate more work in each view.

I realize I came off as defensive. Didn't mean to. Just wanted to document some background for posterity.

sindresorhus:
Just to be clear, the plan was always to consolidate them. I just didn't want to do it before we knew our exact requirements (which could change after user-testing). And it didn't seem like a priority as it's only less than 100 LOC of duplicated simple JSX.

lukechilds:
>There are some other minor differences, like showing the year in the timestamp for the Trades view, not showing the Cancel button in the Exchange view, etc.

We should definitely show the cancel button in all views. Ideally the same date too, but I know space is an issue.

We did talk about removing the "view" button and making the entire row clickable to open the modal. That would buy some space.

sindresorhus:
Yup, I'll experiment with the row clicking and I'll consolidate it soon.

lukechilds:
Perfect!

PR #193, opened 2018-05-09
Format order book API response

### LINKED ISSUE #188: Remove 0 entries from order book
Some entries in the order book have 0 values. They are confusing and it's unclear whether they can be filled or not. We should remove these so we don't confuse users and they aren't included in price calculations.

lukechilds:
Resolves #188
Also makes a start on https://github.com/lukechilds/hyperdex/issues/186

I've made the API reformatting intentionally verbose so you can cleary see what properties are available.

The new formatting:
- Removes all orders with `numutxos` ≤ 0 to increase order book accuracy.
- Removes all time stamps to prevent unnecessary re-render diffs.
- Removes superfluous properties like `numasks`, `numbids`, `askdepth`, `biddepth`.

We now have much less confusing order books with higher accuracy. The price selection buttons and depth chart should also be more accurate.

We can drop the nasty `removeOrderBookTimes()` hack when diffing before setState because I'm no longer including any timestamps.

Properties like `numasks`, `numbids`, `askdepth`, `biddepth` were pointless because all of that data can be easily calculated from the data in the `bids` and `asks` array.

So now instead of:

```jsx
<DepthChart
  bids={orderBook.bids}
  asks={orderBook.asks}
  bidDepth={orderBook.biddepth}
  askDepth={orderBook.askdepth}
/>
```

We've just got:

```jsx
<DepthChart bids={orderBook.bids} asks={orderBook.asks}/>
```

Which makes more sense to me.


lukechilds:
I'm also open to renaming some more of the `order` properties if you have any suggestions.

app/renderer/api.js
```diff
@@ -110,12 +110,33 @@ export default class Api {
...
 			rel,
 		});
+
+		const formatOrders = orders => orders
+			.filter(order => order.numutxos > 0)
+			.map(order => ({
+				address: order.address,
+				depth: order.depth,
+				price: order.price,
+				numutxos: order.numutxos,
```
sindresorhus: `utxoCount`

app/renderer/api.js
```diff
@@ -110,12 +110,33 @@ export default class Api {
...
+				avevolume: order.avevolume,
```
sindresorhus: `averageVolume`

app/renderer/api.js
```diff
@@ -110,12 +110,33 @@ export default class Api {
...
+				maxvolume: order.maxvolume,
```
sindresorhus: `maxVolume`

app/renderer/api.js
```diff
@@ -110,12 +110,33 @@ export default class Api {
...
+				zcredits: order.zcredits,
```
sindresorhus: `zCredits`

What is even this?

sindresorhus:
Looks great!

app/renderer/api.js
lukechilds: It's to do with 0-conf payments. We can use our own name but would prefer to leave that untill I have a better understanding of how it all works.

PR #196, opened 2018-05-09
Show swap error codes

### LINKED ISSUE #194: Show errors in the swap modal
Just the error code for now.

But should show descriptive error messages as we get more info on them.

Complete list here: https://docs.google.com/spreadsheets/d/1blgmrJCZ-SfYaJdTTVdk717kZ43R4yzm9YIW2Bt--EM/edit#gid=0

lukechilds:
Resolves #194
Also makes a start on https://github.com/lukechilds/hyperdex/issues/194

This PR adds a new `error` property to the swap object. By default it is false so we can just do simple truthy checks on it.

If there is an error it will look like this:

```js
swap.error = {
	code: Number,
	message: String || undefined,
}
```

Currently messages are always undefined but this will change as we get more [documentation from James](https://docs.google.com/spreadsheets/d/1blgmrJCZ-SfYaJdTTVdk717kZ43R4yzm9YIW2Bt--EM/edit?usp=sharing)

I'm just displaying the code in the swap details modal but we should display a helpful message if available.

I've just added it to the modal in a basic way, feel free to change it  or style it differently.

screen shot 2018-05-09 at 1 28 49 pm

### DIFF
--- a/app/renderer/swap-db.js
+++ b/app/renderer/swap-db.js
@@ -145,8 +145,7 @@ class SwapDB {
 				// TODO: Add error messages once we have errors documented
 				// https://github.com/lukechilds/hyperdex/issues/180
 				swap.error = {
-					code: message.error,
-					message: undefined,
+					message: `Error Code: ${message.error}`,
 				};
 			}
 
--- a/app/renderer/swap-db.js
+++ b/app/renderer/swap-db.js
@@ -74,6 +74,7 @@ class SwapDB {
 			timeStarted,
 			status: 'pending',
 			statusFormatted: 'pending',
+			error: false,
 			progress: 0,
 			baseCurrency: response.base,
 			quoteCurrency: response.rel,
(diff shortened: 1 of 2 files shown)


app/renderer/swap-db.js
```diff
@@ -140,6 +141,13 @@ class SwapDB {
 
 			if (message.method === 'failed') {
 				swap.status = 'failed';
+
+				// TODO: Add error messages once we have errors documented
+				// https://github.com/lukechilds/hyperdex/issues/180
+				swap.error = {
+					code: message.error,
+					message: undefined,
```
sindresorhus: I don't think the UI should be concerned with when to show code or message. It should be handled here, so that if there's no `message`, we just use `Error Code: -231` as message. That way the consumer code doesn't need to change when the actually get messages for something things.

The below could then be changed to just:

```js
<p>
	{title(swap.statusFormatted)}
	{(swap.status === 'failed' && swap.error) && (
		<React.Fragment>
			{swap.error.message}
		</React.Fragment>
	)}
</p>
```

lukechilds: Good point, will update.

lukechilds:
I've addresses your feedback. I also removed the `code` property so we only have message, is that what you meant?

Now I'm thinking it would be good to have the code available too.

sindresorhus:
> Now I'm thinking it would be good to have the code directly available too.

I think the code should still be there as a property.

lukechilds:
@sindresorhus fixed

PR #199, opened 2018-05-10
Consolidate all supported currency data in one place

### LINKED ISSUE #185: Refactor currency code to make it easier to add new currencies.
We should have a single collection of all supported currencies and any parameters they require throughout the app.

Supporting a new currency in the GUI should be as simple as adding the parameters to this single file.

lukechilds:
Resolves #185 

Now all we need to do is update `app/marketmaker/supported-coins.js` to support new coins. Everything is consolidated there.

I've also removed all the coin data for coins we aren't using (maybe we'll get a `marketmaker` speedup).

Also allowed me to remove some of the dirty custom JSON handling we were having to do.

I think this is loads neater and looks a lot easier to manage. But maybe:

```js
import supportedCoins from '../marketmaker/supported-coins';
```

from inside the `renderer/api.js` folder is a bit dirty? What do you think?


app/renderer/api.js
```diff
@@ -2,7 +2,7 @@ import util from 'util';
 import electron from 'electron';
 import {sha256} from 'crypto-hash';
 import PQueue from 'p-queue';
-import electrumServers from './electrum-servers';
+import supportedCoins from '../marketmaker/supported-coins';
```
sindresorhus: I think we should standardize on the word `currency` and not `coin`. So `supported-currencies`. What do you think?

lukechilds:
The pattern I'm following to make it more readable is:

```js
{
  coin: 'DOGE',																																// <- Ticker
  rpcport: 22555, pubtype: 30, p2shtype: 22, wiftype: 158, txfee: 100000000,	// <- Noisy marketmaker required args
  electrumServers: [																													// <- The electrum servers we use
    {host: '173.212.225.176', port: 10060},
    {host: '136.243.45.140', port: 10060},
  ],
},
```

app/marketmaker/supported-coins.js
```diff
@@ -0,0 +1,68 @@
...
+		coin: 'SUPERNET',
+		asset: 'SUPERNET', rpcport: 11341,
+		electrumServers: [
+			{host: 'electrum1.cipig.net', port: 10005},
+			{host: 'electrum2.cipig.net', port: 10005},
+		],
+	},
+	{
+		coin: 'CHIPS',
+		rpcport: 57776, pubtype: 60, p2shtype: 85, wiftype: 188, txfee: 10000,
```
sindresorhus: Newlines please

app/marketmaker/supported-coins.js
```diff
@@ -0,0 +1,68 @@
...
+		electrumServers: [
+			{host: 'electrum1.cipig.net', port: 10053},
```
sindresorhus: Same here

app/marketmaker/supported-coins.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
+
+const supportedCoins = [
+	{
+		coin: 'KMD',
```
sindresorhus: `coin` => `symbol`

app/marketmaker/supported-coins.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
+
```
sindresorhus: Although more verbose, this would be better as a `Map`. We only ever need to get an item by `symbol`.

```js
const supportedCoins = new Map([
	[
		'KMD',
		{
			electrumServers: [
				{host: 'electrum1.cipig.net', port: 10001},
				{host: 'electrum2.cipig.net', port: 10001},
			],
		},
	],
]);
```

If you don't like the verboseness, you could just keep it as an array and just iterate through it to make it a `Map` before exporting it.

app/renderer/api.js
lukechilds: 👍

app/marketmaker/supported-coins.js
lukechilds: Thought you might have a problem with that ;)

I think it makes it loads easier to skim through and have all the marketmaker fluff on one line like this.

Fixed in https://github.com/lukechilds/hyperdex/pull/199/commits/c6fae3e91836a46ade38b16cc4820a4faa6ef2fb.

app/marketmaker/supported-coins.js
lukechilds: I would prefer this too but we need this for `markermaker` to understand it.

app/marketmaker/supported-coins.js
lukechilds: I would prefer this too but we need this for `markermaker` to understand it.

lukechilds:
We can't change `coins` => `symbol` or change to a map because this is converted to JSON and passed directly as a CLI ark to `marketmaker`.

We could have our own formatted file and then a function that transforms in into `marketmaker` compatible syntax. That seems a bit over the top though, what do you think?

app/renderer/api.js
lukechilds: Fixed https://github.com/lukechilds/hyperdex/pull/199/commits/9bcd2ec7714390d156e3bf6dc182d5d05cedb6ed

lukechilds:
New currency formatting:

```js
{
  coin: 'DOGE',
  rpcport: 22555,
  pubtype: 30,
  p2shtype: 22,
  wiftype: 158,
  txfee: 100000000,
  electrumServers: [
    {
      host: '173.212.225.176',
      port: 10060
    },
    {
      host: '136.243.45.140',
      port: 10060
    },
  ],
},
```

app/marketmaker/supported-currencies.js
```diff
@@ -0,0 +1,134 @@
...
+			},
+		],
+	},
+	{
+		coin: 'CHIPS',
+		rpcport: 57776,
+		pubtype: 60,
+		p2shtype: 85,
+		wiftype: 188,
+		txfee: 10000,
```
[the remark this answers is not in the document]
lukechilds: I'm guessing it means we can add Bitcoin protocol coins ourselves too if we have the coin params. Which is one good reason to keep it seperated.

lukechilds:
Oh nice, look what I just found!

https://github.com/jl777/coins/blob/master/coins

PR #202, opened 2018-05-10
Add ability to toggle currencies

### LINKED ISSUE #130: Allow user to enable/disable currencies
Currently there's no way for the user to enable/disable coins via the UI.

We should enable some popular currencies by default and give the user the option to enable/disable all the currencies we support.

lukechilds:
Resolves #130 

This was a bit more involved than I thought. We have calls to `config.get('enabledCoins')` at different places in the code but it was sometimes stored and treated as a static property.

This meant updating currencies could mean that values on other places using `config.get('enabledCoins')` will become out of sync with the actual enabled coins. So I've moved `enabledCoins` into `appContainer.state` and now everything references it from there.

We should also probably refactor to use `appContainer.state.enabledCoins` everywhere in the app instead of `appContainer.state.currencies` which we poll `marketmaker` for.

At the moment, because most places use the `marketmaker` response if you enable a new currency and then switch to the dashboard view, there can be a delay of a few seconds and then suddenly your new currency will appear. This doesn't make sense because we know which currencies are enabled immediately.

This PR is already getting a bit heavy and it works as is, it's just not optimal. So the refactor should probably be in a follow up PR.

This needs a bit more testing before it's ready to merge. It'll break stuff if you disable a currency that's in the current trading pair.

Also, the UI needs a lot of 💄

toggle-currencies

### DIFF
--- a/app/renderer/containers/App.js
+++ b/app/renderer/containers/App.js
@@ -7,6 +7,7 @@ import roundTo from 'round-to';
 import {Container} from 'unstated';
 import loginContainer from 'containers/Login';
 import {appViews} from '../../constants';
+import supportedCurrencies from '../../marketmaker/supported-currencies';
 import fireEvery from '../fire-every';
 import {formatCurrency} from '../util';
 
@@ -69,7 +70,7 @@ class AppContainer extends Container {
 		const FIVE_MINUTES = 1000 * 60 * 5;
 
 		await fireEvery(async () => {
-			this.coinPrices = await Promise.all(this.enabledCoins.map(symbol => getTickerData(symbol)));
+			this.coinPrices = await Promise.all(supportedCurrencies.map(currency => getTickerData(currency.coin)));
 		}, FIVE_MINUTES);
 	}
 
--- a/app/renderer/containers/App.js
+++ b/app/renderer/containers/App.js
@@ -34,12 +34,12 @@ const getTickerData = async symbol => {
 class AppContainer extends Container {
 	state = {
 		activeView: 'Login',
+		enabledCoins: config.get('enabledCoins'),
 	};
 
 	constructor() {
 		super();
 		this.views = new Cycled(appViews);
-		this.enabledCoins = config.get('enabledCoins');
 
 		this.getSwapDB = new Promise(resolve => {
 			this.setSwapDB = resolve;
@@ -119,6 +119,24 @@ class AppContainer extends Container {
 
 	getCurrency(symbol) {
 		return this.state.currencies.find(x => x.coin === symbol);
+	}
+
+	enableCoin(coin) {
+		this.setState(prevState => {
+			this.api.enableCoin(coin);
+			const enabledCoins = [...prevState.enabledCoins, coin];
+			config.set('enabledCoins', enabledCoins);
+			return {enabledCoins};
+		});
+	}
+
+	disableCoin(coin) {
+		this.setState(prevState => {
+			this.api.disableCoin(coin);
+			const enabledCoins = prevState.enabledCoins.filter(enabledCoin => enabledCoin !== coin);
+			config.set('enabledCoins', enabledCoins);
+			return {enabledCoins};
+		});
 	}
 
 	logOut() {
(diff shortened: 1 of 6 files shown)


sindresorhus:
Nice!

sindresorhus:
If I select BTC in the Exchange view, then disable BTC in the settings, then go back to the Exchange, it crashes:

```
index.js:29 TypeError: Cannot read property 'filter' of undefined
    at formatOrders (api.js:130)
    at api.js:145
    at Generator.next (<anonymous>)
    at step (index.js:11)
    at index.js:11
    at <anonymous>
```

app/renderer/containers/App.js
```diff
@@ -120,6 +121,24 @@ class AppContainer extends Container {
 		return this.state.currencies.find(x => x.coin === symbol);
 	}
 
+	enableCoin(coin) {
```
sindresorhus: For consistency, I would use `enableCurrency(symbol)` here.

lukechilds:
>If I select BTC in the Exchange view, then disable BTC in the settings, then go back to the Exchange, it crashes:

Yup 👇 

>This needs a bit more testing before it's ready to merge. It'll break stuff if you disable a currency that's in the current trading pair.

app/renderer/containers/App.js
lukechilds: It's consistent with the current `enableCoin` api method.

But I agree, I really need to go through the entire app and change all terminology to `coin => currency`.

Probably better suited for a different PR though.

It's also reaaally confusing when you're switching between mm docs and our own code. They use coin instead of currency and `coin.coin` instead of `coin.symbol`.

sindresorhus: Sure

Yeah, and `coin.coin` is super confusing.

lukechilds:
>We should also probably refactor to use appContainer.state.enabledCoins everywhere in the app instead of appContainer.state.currencies which we poll marketmaker for.

On second thoughts I don't think this is a good idea. We might enable a coin, show it somewhere instantly, but then the response might come back from `marketmaker` a few seconds later saying there was a problem enabling it.

We would then have to un enable it again which would be very strange and hard to show a useful error if this could happen after they've navigated away from the view.

Also, that's assuming we actually catch the error, we've got no docs so an error could slip by un noticed.

So even if it's a bit laggy, we're probably better of relying on the `marketmaker` `portfolio` responses as the single source of truth for what currencies are enabled, to ensure consistency between UI <-> `marketmaker`.

sindresorhus:
> So even if it's a bit laggy, we're probably better of relying on the marketmaker portfolio responses as the single source of truth for what currencies are enabled, to ensure consistency between UI <-> marketmaker.

I agree, considering what we're dealing with.

lukechilds:
@sindresorhus fixed, you can now disable coins in the trading pair.

You will still possibly see the

```
index.js:29 TypeError: Cannot read property 'filter' of undefined
    at formatOrders (api.js:130)
    at api.js:145
    at Generator.next (<anonymous>)
    at step (index.js:11)
    at index.js:11
    at <anonymous>
```

error.

This is because the order book doesn't check the active pair in the response is the same as when the request was sent. This causes other issues too. been meaning to fix for a while, will do it in a follow up PR.

Is this good to merge as is?

lukechilds:
@sindresorhus are you able to have a look at https://github.com/lukechilds/hyperdex/pull/202/commits/7920407340d7dc1ca960ae833e1fcc402ebe2ad5 when you get a chance?

I can't understand why it doesn't work. It complains that `appContainer` is undefined but when it get's called later on in other methods it's defined. It's only undefined when the constructor function is called.

I'd understand if it was some `appContainer` property that hasn't been setup yet but how can `appContainer` be undefined and then be defined later on in the same scope?It's imported at the top of the file, how can this be possible? 🤯

sindresorhus:
@lukechilds Solution: https://github.com/lukechilds/hyperdex/pull/207

lukechilds:
@sindresorhus thanks for your fix, this should be working now.

sindresorhus:
Working great now! 🙌

PR #207, opened 2018-05-12
Prevent potential circular dependency

sindresorhus:
Fixes https://github.com/lukechilds/hyperdex/pull/202#issuecomment-388318577

The problem was caused by a circular dependency. This should usually just be handled by the system, but Webpack is of course not able to...

`Containers/App` imports `Containers/Login` which imports `Containers/Dashboard`. It becomes a circular dependency then when we try to import `Containers/Dashboard` in `Containers/App`.

It still seems like it should have always been `undefined` then, right? Since variables cannot suddenly change their value. The thing is, Webpack transforms all imports, so this:

```js
console.log(appContainer);
```

Is transformed into this:

```js
console.log(containers_App__WEBPACK_IMPORTED_MODULE_3__["default"]);
```

And now it's suddenly an object property, which *can* change dynamically.

Gotta love modern web development...

### DIFF
--- a/app/renderer/containers/Login.js
+++ b/app/renderer/containers/Login.js
@@ -6,7 +6,6 @@ import {minWindowSize} from '../../constants';
 import Api from '../api';
 import SwapDB from '../swap-db';
 import appContainer from './App';
-import dashboardContainer from './Dashboard';
 
 const config = remote.require('./config');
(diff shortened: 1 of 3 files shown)


lukechilds:
I had a hunch it was something build related, but I was expecting to be able to track it down with console.logs.

It makes sense now, knowing that the imports have been changed to properties. Good detective work!

PR #208, opened 2018-05-12
Reset state on logout

### LINKED ISSUE #206: Ensure all state is completely reset on user logout
This includes killing the mm process, all container instances, and the balance/address cache.

sindresorhus:
Fixes #206

Reloading the window considerably reduces the chance of data leak and also resolves any memory leaks.

It looks pretty good except for a small flash of lighter background. That is caused by https://github.com/electron/electron/issues/9671 and will hopefully be fixed in the future.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -78,6 +78,7 @@ function createMainWindow() {
 		maximizable: false,
 		fullscreenable: false,
 		titleBarStyle: 'hiddenInset',
+		backgroundColor: '#1b232f', // Same as `--background-color`
 		darkTheme: isDarkMode, // GTK+3
 		webPreferences: {
 			blinkFeatures: 'CSSBackdropFilter',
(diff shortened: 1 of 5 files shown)


lukechilds:
This could maybe be made a bit nicer by adding a fade-in affect to the initial page load.

It would make the background difference less noticeable and maybe also look a bit smoother.

Thoughts?

lukechilds:
Example:

```diff
diff --git a/app/renderer/views/Login.scss b/app/renderer/views/Login.scss
index 06ac5b07..594ff866 100644
--- a/app/renderer/views/Login.scss
+++ b/app/renderer/views/Login.scss
@@ -1,4 +1,12 @@
+@keyframes Login-fade-in {
+	to {
+		opacity: 1;
+	}
+}
+
 .Login {
+	opacity: 0;
+	animation: Login-fade-in 1s ease-in forwards 0.05s;
 	text-align: center;
 	display: flex;
 	justify-content: center;
```

sindresorhus:
@lukechilds Awesome idea! It's much smoother now.

PR #209, opened 2018-05-14
Follow Electron security best practices

### LINKED ISSUE #10: Ensure we follow security best practices
We need to make sure we follow the best practices for security before the public launch: https://github.com/electron/electron/blob/master/docs/tutorial/security.md#checklist

sindresorhus:
This is the strictest CSP possible.

Fixes #10 

CSP cheatsheet: https://content-security-policy.com

---

When/if https://github.com/electron/electron/issues/12909 gets resolved, we can make it even stricter, by limiting it to the local marketmaker port we've selected. We can also remove the `dev.html` then.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -89,7 +89,7 @@ function createMainWindow() {
 	});
 
 	if (is.development) {
-		win.loadURL('http://localhost:8080');
+		win.loadURL('http://localhost:8080/dev.html');
 	} else {
 		loadUrl(win);
 	}
--- a/app/renderer/dev.html
+++ b/app/renderer/dev.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<!-- This file is needed as we can't use CSP in development because hot-reloading requires eval and other things -->
+<html>
+	<head>
+		<title>HyperDEX</title>
+		<meta charset="utf-8">
+		<script defer src="/bundle.js"></script>
+	</head>
+	<body>
+		<div id="root"></div>
+	</body>
+</html>
--- a/app/renderer/index.html
+++ b/app/renderer/index.html
@@ -3,6 +3,7 @@
 	<head>
 		<title>HyperDEX</title>
 		<meta charset="utf-8">
+		<meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'self'; img-src 'self' data:; style-src 'self' 'unsafe-inline'; font-src 'self'; connect-src 'self' http://127.0.0.1:* ws://127.0.0.1:* https://api.coinmarketcap.com https://min-api.cryptocompare.com;">
 		<script defer src="/bundle.js"></script>
 	</head>
 	<body>


lukechilds:
>Fixes #10

Is this the only thing we need to fix, does our boilerplate already handle the other points?

>1. Only load secure content
>1. Disable the Node.js integration in all renderers that display remote content
>1. Enable context isolation in all renderers that display remote content
>1. Use ses.setPermissionRequestHandler() in all sessions that load remote content
>1. Do not disable webSecurity
>1. Define a Content-Security-Policy and use restrictive rules (i.e. script-src 'self')
>1. Override and disable eval, which allows strings to be executed as code.
>1. Do not set allowRunningInsecureContent to true
>1. Do not enable experimental features
>1. Do not use blinkFeatures
>1. WebViews: Do not use allowpopups
>1. WebViews: Verify the options and params of all <webview> tags
```

**Comment by sindresorhus** [comment]
>1. Only load secure content

CSP enforces this.

>1. Disable the Node.js integration in all renderers that display remote content

Doesn't apply. We don't load remote executable content.

>1. Enable context isolation in all renderers that display remote content

Doesn't apply. We don't load remote executable content.

>1. Use ses.setPermissionRequestHandler() in all sessions that load remote content

Doesn't apply. We don't load remote executable content.

>1. Do not disable webSecurity

We don't disable this.

>1. Define a Content-Security-Policy and use restrictive rules (i.e. script-src 'self')

Done in this PR. CSP handles most security concerns.

>1. Override and disable eval, which allows strings to be executed as code.

CSP handles this.

>1. Do not set allowRunningInsecureContent to true

We don't set this.

>1. Do not enable experimental features

We don't set this.

>1. Do not use blinkFeatures

We only set one feature and it's not security related. It's only for the modal blurring.

>1. WebViews: Do not use allowpopups

We don't use webviews.

>1. WebViews: Verify the options and params of all <webview> tags

We don't use webviews.

**Comment by sindresorhus** [comment]
Our biggest security issue is marketmaker, but that's outside what we can handle in Electron.

**Comment by lukechilds** [comment]
Awesome!

**Comment by lukechilds** [comment]
>Our biggest security issue is marketmaker

And dependency management.

**Comment by MarshallOfSound** [comment]
> Doesn't apply. We don't load remote executable content.

Personal recommendation, do it anyway, I don't know exactly what this app does but an XSS / RCE becomes way less potent with these measures in place.

> We don't use webviews.

Disable them explicitly then  `webviewTag: false` in your `webPreferences`

**Comment by MarshallOfSound** [comment]
> We can't disable Node.js integration as we use that heavily.

If we're going down the route of "let's make a super secure and safe Electron app" then you shouldn't be using `nodeIntegration` in the renderer, you should be using it in a `preload` script with context isolation enabled.

Anything that requires `nodeIntegration` should be sent over `message` events to and from the `preload` script.  This removes `nodeIntegration` from the playpen if an XSS is found in your app and normally has the added bonus of a nicer modular app layout 😆 

> What does it do that CSP doesn't already handle?

CSP *should* prevent a lot of things, I'm saying follow all these other security measures Just In Case  it doesn't.  If someone finds a way around your CSP then the response shouldn't be "the CSP should have stopped them", having additional layers of protection can only ever be a good thing.

In this case setting the permission handler will mean if anyone finds an XSS and manages to execute arbitrary JS they won't be able to get access to the user webcam for instance 😄 

_this is all just IMO security practices_

**Comment by lukechilds** [comment]
This is a good point re Node.js integration, currently if someone managed to craft some kind of strange order entry or whatever that somehow could inject JS (I know ideally they shouldn't be able to do this in the first place) all they need to do is:

```js
require('electron').remote.require('child_process').execSync('whoami').toString();
```

And they've got total control of the machine. They could easily connect to a reverse shell, install a persistent back door, steal funds etc.

liliakai:
> https://content-security-policy.com

Just want to point out that this is website is  based on CSP Level 1 and 2, and may not represent the latest best practice. For instance, it declares `frame-src` as deprecated in favor of `child-src`, a decision which has been reversed in CSP Level 3.

In contrast, this MDN page recommends use of separate `frame-src` and `worker-src` rather than `child-src`: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy

✌️

lukechilds:
Thanks for the heads up @liliakai, we'll look in to that ❤️

sindresorhus:
@liliakai That's good to know. I should have known we could only trust MDN. We don't use `frame-src`, so it doesn't really apply for us.

PR #215, opened 2018-05-14
Encrypt swap database

### LINKED ISSUE #132: Swap DB should be encrypted
If someone else gets access to your backup they shouldn't be able to view swap info. It contains TXIDs which someone could easily derive the users addresses and balance from.

lukechilds:
Resolves #132 

Swap database is now encrypted using the seed phrase (which is passed through PBKDF2 internally).

**Important:** This completely breaks existing swap databases (because they can't be decrypted). You'll need to delete your DB.

To migrate your swap DB do the following:

```js
// Dump your current DB and save it somewhere
_swapDB._getAllSwapData().then(JSON.stringify).then(console.log);

// Destroy your current DB
_swapDB.db.destroy();

// Checkout this branch and restart HyperDEX

// Import your old swaps into the new DB
dbDumpYouSavedEarlier.forEach(swap => {
	delete swap._id;
	delete swap._rev;
	_swapDB.db.post(swap);
});
```

### DIFF
--- a/app/renderer/containers/Login.js
+++ b/app/renderer/containers/Login.js
@@ -79,11 +79,11 @@ class LoginContainer extends Container {
 	async handleLogin(portfolioId, password) {
 		const portfolio = this.portfolioFromId(portfolioId);
 
-		const swapDB = new SwapDB(portfolioId);
-		appContainer.setSwapDB(swapDB);
-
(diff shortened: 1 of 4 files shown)


lukechilds:
It would be possible to implement some kind of auto upgrade code but I don't think we should bloat the code with that.

Testers should follow these steps to migrate or just nuke their DB.

PR #218, opened 2018-05-14
Add initial Swap view styling

sindresorhus:
It bothered me that it looked so unfinished and ugly. I just whipped this up in 10 min.

@lukechilds Any suggestion for improvements to the text?

screen shot 2018-05-14 at 18 24 32


lukechilds:
Good idea, how about:

>Simple atomic swaps with minimal user interaction

PR #219, opened 2018-05-15
Improve the currency selection option

sindresorhus:
It supports selecting or searching for a currency to enable. You can search and press enter to quickly enable. You can press backspace to delete the previous addition. You can press the `x` to delete previous additions. KMD and CHIPS cannot be deleted.

### DIFF
--- a/app/renderer/components/Select.scss
+++ b/app/renderer/components/Select.scss
@@ -212,7 +212,9 @@
 
 .Select--multi .Select-multi-value-wrapper {
 	display: inline-block;
-	padding-bottom: 6px;
+	padding: 4px 2px 8px;
+	max-height: 100px;
+	overflow-y: auto;
 }
 
 .Select .Select-aria-only {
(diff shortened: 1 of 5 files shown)


lukechilds:
I really like it but I'm not sure how usable it's going to be with a large amount of coins. This is only with 9 coins enabled and it already becomes a bit unclear that it's a select input:

screen shot 2018-05-15 at 2 44 30 pm

If a user had 20 or 30 coins enabled the drop down selection might be off the visible page.

One potential solution could be to keep the select component but not list all the currencies inside it, list them below instead. What do you think?

sindresorhus:
> If a user had 20 or 30 coins enabled the drop down selection might be off the visible page.

That's a good point. I've set a max height on it now, so that it scrolls the content if there are a lot of currencies. I've set it to a low number so you can test. I'll increase it if accepted so that it doesn't need scrolling quite as early.

lukechilds:
That's much better 👌

PR #225, opened 2018-05-17
Add PIZZA and BEER test currencies

sindresorhus:
This was requested by testers.

I also improved the error handling of the ticker data fetching.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -1,24 +1,29 @@
 'use strict';
 const Store = require('electron-store');
 const {minWindowSize} = require('./constants');
+const {isNightlyBuild} = require('./util-common');
 
-module.exports = new Store({
-	defaults: {
-		windowState: {
-			width: minWindowSize.width,
-			height: minWindowSize.height,
-		},
-		darkMode: true,
-		enabledCoins: [
-			'BTC',
-			'KMD',
-			'CHIPS',
-			'SUPERNET',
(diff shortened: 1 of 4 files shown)


lukechilds:
Also

>I also improved the error handling of the ticker data fetching.

I think this should be in a separate PR.

sindresorhus:
> I think this should be in a separate PR.

Yeah, you're right. I was just lazy. It's needed for this PR as I noticed it was breaking when I added PIZZA.

sindresorhus:
> ETOMIC is only needed for ETH/ERC20 swaps, it's kind of like a proxy token to enable the atomic swap as ETH doesn't have UTXOs. We don't support ETH/ERC20 yet and it might be a while till we do. Artem also talked about making the requirement of ETOMIC completely transparent, so this might end up being a special case where we hide this from the UI.

Ok. I didn't know that. I thought it was just another test coin. Removed.

> PIZZA and BEER are test asset chains for stress testing the BarterDEX network and are given away for free. It's a good idea to enable these but I think they should only be enabled in dev/nightly builds. It might be confusing for normal users. What do you think?

I agree. Done: https://github.com/lukechilds/hyperdex/pull/225/commits/d7f22c1093c8372bc67a196a0830bdb4adb8002b

You can still add them manually in the preferences file for testing the stable build in production.

PR #226, opened 2018-05-17
Work around React issue with number inputs and zero

sindresorhus:
Doesn't seem like https://github.com/facebook/react/issues/9402 will be fixed any time soon, so we'll work around it for now...


lukechilds:
How does this work exactly?

Just by forcing the value to 0 and then setting a real value?

sindresorhus:
> Just by forcing the value to 0 and then setting a real value?

Yes, the React issue is that React cannot detect the difference between `01` and `1`. The workaround just forces `0` first, so React can notice the difference.

PR #227, opened 2018-05-17
Improve text size for large resolutions

### LINKED ISSUE #176: Improve text size
Some people have commented some of the text size is too small in some places in the Exchange view. We have a space problem there, but should be able to slightly increase the text size.

Related: https://github.com/lukechilds/hyperdex-bugtracker/issues/4

sindresorhus:
Fixes #176
Fixes lukechilds/hyperdex-bugtracker#4

When the window is 1360px or wider, the order columns are made bigger and some of the smaller text in the Exchange view is made larger.

### DIFF
--- a/app/renderer/views/Exchange/Exchange.scss
+++ b/app/renderer/views/Exchange/Exchange.scss
@@ -42,4 +42,8 @@
 		background-color: var(--widget-background-color);
 		border-radius: 4px;
 	}
+
+	@media (min-width: 1360px) {
+		grid-template-columns: 320px 1fr 320px;
+	}
 }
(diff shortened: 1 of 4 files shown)


lukechilds:
The text size increase should be done for all resolutions, not just larger ones, the issue isn't caused by high resolutions, it's caused by high DPI, which will bo noticeable at all sizes.

If you set your MacBook to 1680x1050, which is the default for 15" MacBooks the order book text is quite small. If you set it to 1920x1200 it's very small. Even if you have the window opened <1360px wide, the text will still appear small.

I think making the order columns wider over a certain width makes sense though, it makes better use of the extra space.

Also, I think the swap history text is ok as it is, there's less whitespace around them with the larger order columns.

sindresorhus:
I've addressed your feedback. Gonna merge this now to get it into the release, but comment here whenever if I missed anything.

PR #232, opened 2018-05-18
Update exchange intro text

### LINKED ISSUE #223: Update first run modal text with correct information
The current text in the first run modal on the exchange view has incorrect information, we should update this before launch.

lukechilds:
Resolves #223 

Updated the text intro to have correct information. Also linked to some external references.

Also added an `<ExternalLink>` component.

screen shot 2018-05-18 at 3 07 25 pm

### DIFF
--- a/app/renderer/components/ExternalLink.js
+++ b/app/renderer/components/ExternalLink.js
@@ -1,8 +1,8 @@
 import electron from 'electron';
 import React from 'react';
 
-const open = electron.shell.openExternal;
+const {openExternal} = electron.shell;
 
-const ExternalLink = ({url, ...props}) => <a {...props} onClick={() => open(url)}/>;
+const ExternalLink = ({url, ...props}) => <a {...props} onClick={() => openExternal(url)}/>;
 
 export default ExternalLink;
--- a/app/renderer/components/ExternalLink.js
+++ b/app/renderer/components/ExternalLink.js
@@ -1,8 +1,9 @@
 import electron from 'electron';
 import React from 'react';
+import Link from './Link';
 
 const {openExternal} = electron.shell;
 
-const ExternalLink = ({url, ...props}) => <a {...props} onClick={() => openExternal(url)}/>;
+const ExternalLink = ({url, ...props}) => <Link {...props} onClick={() => openExternal(url)}/>;
 
 export default ExternalLink;
--- a/app/renderer/components/ExternalLink.js
+++ b/app/renderer/components/ExternalLink.js
@@ -1,5 +1,5 @@
-import React from 'react';
 import electron from 'electron';
+import React from 'react';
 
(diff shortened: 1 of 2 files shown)


app/renderer/components/ExternalLink.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import electron from 'electron';
+
+const open = electron.shell.openExternal;
```
sindresorhus: This is shorter and more readable:

```js
const {openExternal} = electron.shell;
```

`open` is too generic.

lukechilds: 👍

app/renderer/views/Exchange/Intro.js
```diff
@@ -18,14 +19,15 @@ class Intro extends React.Component {
 		return (
 			<div>
 				<Modal
-					title="How order books work with HyperDEX"
-					open={this.state.shouldOpen}
+					title="How trading works with HyperDEX"
```
sindresorhus: I think we should titlecapitalize words:

```
How Trading Works with HyperDEX
```

sindresorhus:
`on chain` => `on-chain`

It looks weird only linking `UTXO` in `UTXOs`. You should link the whole word.

app/renderer/views/Exchange/Intro.js
lukechilds: Yeah, definitely, done 👍

sindresorhus:
`trading a coin with` => `trading a currency with`?

sindresorhus:
`This means it functions slightly differently to a centralised exchange.` => `This means it works slightly differently than a centralized exchange.`?

lukechilds:
Thanks, good spots.

app/renderer/components/ExternalLink.js
```diff
@@ -0,0 +1,8 @@
+import electron from 'electron';
+import React from 'react';
+
+const {openExternal} = electron.shell;
+
+const ExternalLink = ({url, ...props}) => <a {...props} onClick={() => openExternal(url)}/>;
```
sindresorhus: I think this should use the `Link` component instead, so it get's the `.Link` classname.

lukechilds: I did originally but then it uses a `<button>` which messes up the styles when it's embedded in text. It won't allow any word breaks in the button element.

It makes sense for UI links to be buttons but I don't think it does for textual links in a sentence.

sindresorhus: @lukechilds I've fixed this here: https://github.com/lukechilds/hyperdex/pull/234 It's now an `a` too.

lukechilds:
Yummy

kapture 2018-05-18 at 16 30 21

PR #235, opened 2018-05-18
Upgrade Unstated and await `setState`

sindresorhus:
We can now await `setState` which simplifies a lot of code \o/


lukechilds:
Very nice

PR #244, opened 2018-05-21
Fix depth chart

sindresorhus:
It was previously sorted by price, which worked because the depth was according to price, now that we have more activity, this issue became more apparent. I also now order before rounding the price, so it correctly sorts very low numbers.


lukechilds:
Perfect 👌

PR #245, opened 2018-05-21
Require user confirmation of transaction fees before withdrawal

### LINKED ISSUE #104: Get `txfee` before withdrawing
Before we do a withdraw we should do a request to:

```js
{
  method: 'getfee',
  coin: 'BTC',
}
```

And display the `txfee` to the user in the confirmation dialogue.

We then specify a `txfee` property in satoshis to the withdraw command.

This is a new method I just got James to add for us. It's probably not in a main build yet. And not important right now. Just documenting it here.

lukechilds:
Resolves #104 

Creates a raw transaction and requests the user to confirm they're happy with the network fee before broadcasting it.

This uses a custom build of mm so can't be merged yet. We should have a release from Artem soon containing the fixes we need.

### DIFF
--- a/app/renderer/api.js
+++ b/app/renderer/api.js
@@ -215,7 +215,7 @@ export default class Api {
 		});
 
 		if (!result.complete) {
-			throw errorWithObject('Couldn\'t complete withdrawal', result);
+			throw errorWithObject('Couldn\'t create withdrawal transaction', result);
 		}
 
 		return {
--- a/app/renderer/api.js
+++ b/app/renderer/api.js
@@ -224,10 +224,10 @@ export default class Api {
 		};
 	}
 
-	async broadcastTransaction(coin, transaction) {
+	async broadcastTransaction(currencySymbol, transaction) {
 		const response = await this.request({
 			method: 'sendrawtransaction',
-			coin,
+			coin: currencySymbol,
 			signedtx: transaction,
 		});
 
--- a/app/renderer/api.js
+++ b/app/renderer/api.js
@@ -193,7 +193,7 @@ export default class Api {
 		return response.txfee;
 	}
 
-	async withdraw(opts) {
+	async createTransaction(opts) {
 		if (typeof opts.currency !== 'string') {
 			throw new TypeError(`opts.currency must be a string: ${opts.currency}`);
 		}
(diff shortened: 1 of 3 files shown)


app/renderer/api.js
```diff
@@ -210,14 +210,53 @@ export default class Api {
...
+			amount,
+			currency,
+			address,
+		} = await this.createTransaction(opts);
+
+		// Convert from satoshis
+		const txFee = txFeeSatoshis / 100000000;
+
+		const broadcast = async () => {
+			// This is needed until a bug in marketmaker is resolved
```
sindresorhus: Can you comment a link to the marketmaker issue here?

app/renderer/views/Dashboard/WithdrawModal.js
```diff
@@ -13,6 +13,9 @@ const getInitialProps = () => ({
 	recipientAddress: '',
 	amount: 0,
 	isWithdrawing: false,
+	isBroadcasting: false,
+	txFee: 0,
+	broadcast: false,
```
sindresorhus: `isBroadcast`?

sindresorhus: `broadcast` => `hasBroadcast`

app/renderer/api.js
lukechilds: It will be resolved before I merge this

app/renderer/api.js
```diff
@@ -210,14 +210,53 @@ export default class Api {
...
 		}
 
-		return result;
+		return {
+			...opts,
+			...result,
+		};
+	}
+
+	broadcastTransaction(coin, transaction) {
```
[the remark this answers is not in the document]
lukechilds: I thought it would be useful to expose for something else but we don't need t for anything yet.

`coin => currencySymbol`

app/renderer/api.js
```diff
@@ -210,14 +210,53 @@ export default class Api {
...
+			hex: transaction,
+			txfee: txFeeSatoshis,
+			txid,
+			amount,
+			currency,
+			address,
+		} = await this.createTransaction(opts);
+
+		// Convert from satoshis
+		const txFee = txFeeSatoshis / 100000000;
```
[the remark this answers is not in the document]
lukechilds: 👍

app/renderer/views/Dashboard/WithdrawModal.js
lukechilds: `broadcast` is set as `false` by default but will be a function one the transaction has been created.

`this.state.broadcast()` broadcasts the pending transaction.

app/renderer/api.js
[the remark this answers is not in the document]
sindresorhus: > transaction => txid These are different, it's the actual raw transaction in hexadecimal.

I'm just looking for something more explicit than just `transaction`. `transaction` sounds like it's a transaction object of some sort. Maybe `rawTransactionHex` or something then?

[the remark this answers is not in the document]
lukechilds: >I would place this after withdraw() in the file as it's also called after it. Just for readability purposes.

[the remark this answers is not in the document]
sindresorhus: 👍

app/renderer/views/Dashboard/WithdrawModal.js
lukechilds: >Can you at least make it undefined

Yeah.

>Kinda weird to have a method in the state.

I agree. I originally just had it as a property/method of the component but decided it should be in state because setting it should trigger a re-render. What do you think?

sindresorhus:
Looks like the `Remaining balance` is not updated after the txfee is fetched to take it into account.

lukechilds:
Good spot.

PR #251, opened 2018-05-21
Tweak date formats

lukechilds:
This prefers `dd/mm/yy` over `dd.mm.yy` and some other small tweaks.

Also this slightly improves the spacing of the swap data. Previously 4 digit swap values with 8 decimals would overflow larger ticker symbols.

Before:

screen shot 2018-05-21 at 9 52 00 pm

After:

screen shot 2018-05-21 at 9 48 07 pm

### DIFF
--- a/app/renderer/components/TimeSeriesChart.js
+++ b/app/renderer/components/TimeSeriesChart.js
@@ -17,7 +17,7 @@ const CustomTooltip = ({payload}) => {
 };
 
 const resolutionToLabelFormat = new Map([
-	['hour', 'HH:mm:ss'],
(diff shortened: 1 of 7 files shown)


app/renderer/components/TimeSeriesChart.js
```diff
@@ -18,9 +18,9 @@ const CustomTooltip = ({payload}) => {
 
 const resolutionToLabelFormat = new Map([
 	['hour', 'HH:mm:ss'],
```
sindresorhus: We can drop seconds here. It always shows `00` anyway.

PR #292, opened 2018-05-22
Don't show price for test currencies

sindresorhus:
Marketmaker returns a price for PIZZA, but it doesn't really make sense to show it as it's not real.

### DIFF
--- a/app/renderer/containers/App.js
+++ b/app/renderer/containers/App.js
@@ -14,17 +14,18 @@ import {isDevelopment} from '../../util-common';
 
 const config = remote.require('./config');
 
+const excludedTestCurrencies = new Set([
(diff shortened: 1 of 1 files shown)


sindresorhus:
Marketmaker shouldn't return a price in the first place. @lukechilds If you agree, I'll open a Marketmaker issue, but we still want this PR in the meantime regardless.

lukechilds:
Are you sure the price isn't coming from CMC? PIZZA is listed on CMC which is PizzaCoin. An unrelated currency which does have a price.

sindresorhus:
Marketmaker issue: https://github.com/jl777/SuperNET/issues/862

PR #295, opened 2018-05-23
Hide and ignore "Custom Marketmaker" preference

### LINKED ISSUE #259: Hyperdex login issue, possibly related to changing default showing trading pairs.
Windows 8.1

Issue: HyperDex will not allow me to log in.  Stalled at splash screen 'Logging in...'

Downloaded hyperdex.
Created a portfolio and logged in.
Deposited Beer from faucet into portfolio.
Went to preferences and adjusted what default coins I could see listed.
Exited hyperdex.
Reopened hyperdex and put in login information, >problem start.

Additional steps:  After final step above, closed hyperdex, reopened and entered login information. Problem persisted. 

_Below: Current View_
image

      
_Below: Error box that may or may not be related to event. Error box appears after entering portfolio login information, however did not prevent logging into the platform during the first success._ 

image

---

HyperDEX 0.1.0-alpha.1
Electron 2.0.0
win32 6.3.9600
Locale: en-US

sindresorhus:
- It works but doesn't correctly update the GUI and swaps (reported on Slack).
- We don't properly handle errors during login for it.
- We don't validate the contents, so users can just input anything there (fixes #259).
- It should be in the logged out preferences, which we don't have yet.
- The Marketmaker version has to be the exact same as the one we're using, which is unfeasible as it changes all the time.

So we should hide this until we can do it properly.

### DIFF
--- a/app/renderer/containers/Login.js
+++ b/app/renderer/containers/Login.js
@@ -12,7 +12,9 @@ const config = remote.require('./config');
 const {getPortfolios, decryptSeedPhrase} = remote.require('./portfolio-util');
 
 const initApi = async seedPhrase => {
-	let url = config.get('marketmakerUrl');
+	// TODO: Disabled for now until we can properly support custom marketmaker instances
+	/// let url = config.get('marketmakerUrl');
+	let url = false;
 	if (url) {
 		console.log('Using custom marketmaker URL:', url);
 	} else {
--- a/app/renderer/views/Preferences.js
+++ b/app/renderer/views/Preferences.js
@@ -3,7 +3,6 @@ import React from 'react';
 import _ from 'lodash';
 import {Subscribe} from 'unstated';
 import appContainer from 'containers/App';
-import Input from 'components/Input';
 import CurrencySelectOption from 'components/CurrencySelectOption';
 import Select from 'components/Select';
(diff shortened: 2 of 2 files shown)


lukechilds:
Hmmn, maybe it would be better to add an advanced warning.

---

>It works but doesn't correctly update the GUI and swaps (reported on Slack).

I saw that but I just tested, works fine for me and updates the UI properly. I often use it for testing custom builds.

>We don't properly handle errors during login for it.

It should only be used by advanced testers who want to debug with their own mm version. We should handle this better but I don't think it's crucial for advanced testers.

>We don't validate the contents, so users can just input anything there (fixes #259).

Again, we should definitely validate this but I don't think it will be an issue for advanced testers.

>It should be in the logged out preferences, which we don't have yet.

I agree, but as long as we label it as an advanced feature it's nicely accessible for now.

>The Marketmaker version has to be the exact same as the one we're using, which is unfeasible as it changes all the time.

It doesn't have to be the exact version, it just has to be the same or newer, James doesn't tend to add breaking changes. If testers are testing their own mm they should be building the latest mm from dev branch.

sindresorhus:
Won't get to this anytime soon, so closing for now. We have an issue to track this.

PR #298, opened 2018-05-23
Show empty state for depth chart when no data

sindresorhus:
screen shot 2018-05-23 at 14 37 32


lukechilds:
This is great 👌 

I had this written on a todo somewhere but totally forgot about it.

PR #299, opened 2018-05-23
Rename `Preferences` to `Settings`

sindresorhus:
Like in the original Figma designs. I know I was the one advocating for using `Preferences`, but I've changed my mind. It was for consistency with native apps, but they only use `Preferences` for legacy conventions. iOS, for example, uses `Settings` too. And this is not really app preferences anyway, but rather portfolio settings.


lukechilds:
Great, I prefer settings.

PR #300, opened 2018-05-23
Package run script improvements

sindresorhus:
- We now use `yarn start` instead of `yarn dev`. No reason to have both.
- I fixed the issue where the app would launch blank. The fix is to actually wait for the dev server to become ready before launching Electron.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,11 +2,10 @@
 	"scripts": {
 		"postinstall": "electron-builder install-app-deps",
 		"lint": "xo && stylelint 'app/renderer/**/*.scss'",
-		"test": "npm run lint",
-		"start": "webpack && electron app",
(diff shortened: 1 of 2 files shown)


lukechilds:
Awesome.

Man it's gonna take me so long to get used to `yarn start` now. But makes sense 👍

sindresorhus:
If you have to relearn in anyways, I would recommend `alias ys='yarn start'`. It's something you use all the time, so shorten it down.

lukechilds:
I really need to step up my alias game. I have exactly 0 aliases 😮

I use `zsh-users/zsh-autosuggestions` as a replacement for aliases, which works just as well or better most of the time but obviously is not the same thing and has bitten me a few times.

PR #306, opened 2018-05-24
Fix stopping marketmaker on logout

sindresorhus:
When fixing an issue a long time ago (https://github.com/lukechilds/hyperdex/commit/dfda2d944380691b287ab82f0348eb385d556df8), I introduced a bug. The fix then was to pause the queue so we wouldn't get any zombie calls after logging out. The problem is that the API `stop` command also uses the queue, so it was never run. We now only pause the queue after the `stop` command is done.

### DIFF
--- a/app/renderer/api.js
+++ b/app/renderer/api.js
@@ -288,12 +288,10 @@ export default class Api {
 	}
 
 	async stop() {
+		this.queue.clear();
+		await this.request({method: 'stop'});
 		this.queue.pause();
 		this.queue.clear();
-
-		try {
-			await this.request({method: 'stop'});
-		} catch (_) {} // Ignoring the error as `marketmaker` doesn't return a response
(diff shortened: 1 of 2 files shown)


app/renderer/api.js
```diff
@@ -288,12 +288,10 @@ export default class Api {
...
 
 	async stop() {
+		this.queue.clear();
+		await this.request({method: 'stop'});
 		this.queue.pause();
 		this.queue.clear();
-
-		try {
-			await this.request({method: 'stop'});
-		} catch (_) {} // Ignoring the error as `marketmaker` doesn't return a response
```
sindresorhus: The `stop` command no longer throws, so seems like it was fixed in mm at some point.

lukechilds: I see

PR #307, opened 2018-05-24
Rewrite `<Input>` and `<TextArea>` components to be fully controlled

### LINKED ISSUE #237: Cursor jumping to the end on seed confirm page
When creating a new portfolio, on the seed confirm page, after a few words are entered, if an attempt is made to edit the previous words (i.e., entering text in between the already entered words), after entering a single letter, the cursor jumps to the end of the line.
Demo
---

HyperDEX Nightly 0.0.0-536c5dd-18.5.18.1358
Electron 2.0.1
linux 4.13.0-41-generic
Locale: en-GB

sindresorhus:
Since our state updates are async, it creates a race condition where React cannot correctly diff the updates and it results in the cursor being pushed to the end.

See this React issue: https://github.com/facebook/react/issues/955 In short, just another React annoyance.

Fixes #237

### DIFF
--- a/app/renderer/components/Input.js
+++ b/app/renderer/components/Input.js
@@ -2,85 +2,114 @@ import React from 'react';
 import {classNames} from 'react-extras';
 import './Input.scss';
 
-const Input = ({
-	className,
-	level,
-	message,
-	errorMessage,
-	disabled,
-	readOnly,
-	onChange,
-	type = 'text',
-	icon,
-	iconSize,
-	iconName,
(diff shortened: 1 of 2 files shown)


sindresorhus:
Note that https://github.com/lukechilds/hyperdex/commit/26eab29feaeb6a4913f04074e7c15019b24d9e07 overrides this fix for number inputs. So number inputs will not be fixed for this issue until React fixes it. It's impossible to fix both.

PR #308, opened 2018-05-24
Add `Copy Swap Debug Data` button to the swap modal

sindresorhus:
When in development or nightly build.

@lukechilds You requested this some days ago.


lukechilds:
Thanks, this should help a lot with testers.

PR #313, opened 2018-05-25
Disable Buy/Sell buttons while the order is being placed

sindresorhus:
To prevent issues with interrupted swaps.

### DIFF
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -15,6 +15,7 @@ const getInitialState = () => ({
 		biddepth: 0,
 		askdepth: 0,
 	},
+	isSendingOrder: false,
 });
 
 class ExchangeContainer extends Container {
@@ -100,6 +101,24 @@ class ExchangeContainer extends Container {
 
 		return this.stopWatchingOrderBook;
 	}
+
+	setIsSendingOrder = isSendingOrder => {
+		this.setState({isSendingOrder});
+	};
+
+	isActiveOrderPending = () => {
+		// TODO: Remove this when we have Marketmaker v2.
+		// Work around Marketmaker's limitation of only one pending swap at the time.
+		// Sometimes swaps get stuck in pending mode, so we timeout after one minute.
+		const ONE_MINUTE = 1000 * 60;
+		const [activeSwap] = this.state.swapHistory;
+		const isPending = (
+			activeSwap.status === 'pending' &&
+			(Date.now() - activeSwap.timeStarted) < ONE_MINUTE
+		);
+
+		return this.state.isSendingOrder || isPending;
+	};
 }
 
 const exchangeContainer = new ExchangeContainer();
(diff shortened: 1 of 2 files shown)


app/renderer/views/Exchange/Order.js
```diff
@@ -116,22 +118,28 @@ class Bottom extends React.Component {
...
 
-		this.setState({statusMessage: ''});
-
 		const swap = result.pending;
 
 		const swapDB = await appContainer.getSwapDB;
 		swapDB.insertSwapData(swap, requestOpts);
 		api.subscribeToSwap(swap.uuid).on('progress', swapDB.updateSwapData);
+
+		// There's a delay between the swap data coming back and the database being updated.
```
lukechilds: Would:

```js
await swapDB.insertSwapData(swap, requestOpts);
```
work instead of `setTimeout()` here?

app/renderer/containers/Exchange.js
```diff
@@ -100,6 +101,24 @@ class ExchangeContainer extends Container {
...
+
+	isActiveOrderPending = () => {
+		// TODO: Remove this when we have Marketmaker v2.
+		// Work around Marketmaker's limitation of only one pending swap at the time.
+		// Sometimes swaps get stuck in pending mode, so we timeout after one minute.
+		const ONE_MINUTE = 1000 * 60;
+		const [activeSwap] = this.state.swapHistory;
+		const isPending = (
+			activeSwap.status === 'pending' &&
+			(Date.now() - activeSwap.timeStarted) < ONE_MINUTE
```
lukechilds: >we timeout after one minute

Is this just working as a side affect of us polling every second? Because it forces a lot of stuff to re-render which will then trigger a diff when this re-enables after a minute?

If so, I'm not sure that's a good solution. I think the functionality should be self contained. It shouldn't rely on side affects of unrelated code to render properly. It should have it's own timer to reset the state.

sindresorhus:
I've changed my mind. I don't think we should disable the button for a minute while it's pending. It's not very user-friendly to just disable it for that long without an explanation to the user. Instead, it's now only disabled while the order is sent, to prevent the user from doing duplicate orders by accident and prevents confusing `mm`. If the user tries to do an order while one is pending, we show this message from `mm` anyway, which is better:

> Only one pending swap at a time, try again in 53 seconds.

lukechilds:
>I've changed my mind. I don't think we should disable the button for a minute while it's pending. It's not very user-friendly to just disable it for that long without an explanation to the user. Instead, it's now only disabled while the order is sent, to prevent the user from doing duplicate orders by accident and prevents confusing mm.

👍 

>If the user tries to do an order while one is pending, we show this message from mm anyway, which is better:
>
>>Only one pending swap at a time, try again in 53 seconds.

Just FYI, I'm not sure the mm error is actually correct. I think it is actually possible to place multiple orders at the same time, just not on the same trading pair.

PR #329, opened 2018-05-29
Add `alwaysEnabledCurrencies` constant

lukechilds:
This is required for ETH/ERC20, that's not ready yet but this is still an improvement over the current code so I've cherry picked it out from my local ETH branch.

It's nice to have this defined somewhere rather than hardcode it in a few places.

This also ensures these currencies are always activated (even if they're removed from the config). That's good because it'll break things if KMD (and soon ETOMIC) aren't enabled.

This also means we can easily enforce new currencies in updates (like ETOMIC). Updating the default coins in the config won't work after the initial install.

### DIFF
--- a/app/renderer/containers/Login.js
+++ b/app/renderer/containers/Login.js
@@ -2,7 +2,7 @@ import {remote} from 'electron';
 import {setWindowBounds} from 'electron-util';
 import ipc from 'electron-better-ipc';
 import {Container} from 'unstated';
-import {minWindowSize, alwaysEnabledCurrencies} from '../../constants';
+import {minWindowSize} from '../../constants';
 import {isDevelopment} from '../../util-common';
 import Api from '../api';
 import SwapDB from '../swap-db';
--- a/app/renderer/containers/Login.js
+++ b/app/renderer/containers/Login.js
@@ -92,7 +92,6 @@ class LoginContainer extends Container {
 			window._swapDB = swapDB;
 		}
 
-		await Promise.all(alwaysEnabledCurrencies.map(x => api.enableCurrency(x)));
 		await Promise.all(appContainer.state.enabledCoins.map(x => api.enableCurrency(x)));
 
 		await appContainer.watchCMC();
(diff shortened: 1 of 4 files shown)


app/renderer/containers/Login.js
```diff
@@ -92,6 +92,7 @@ class LoginContainer extends Container {
 			window._swapDB = swapDB;
 		}
 
+		await Promise.all(alwaysEnabledCurrencies.map(x => api.enableCoin(x)));
```
sindresorhus: I think this is moot. You've already added the `alwaysEnabledCurrencies` to `appContainer.state.enabledCoins` in https://github.com/lukechilds/hyperdex/pull/329/files#diff-31e11a5a0f9265edd7943561f9166aadR63

lukechilds: You're right, it is currently, but it's required in the ETH branch.

ETOMIC will be added to `alwaysEnabledCurrencies` and needs to be enabled before enabling ETH or any ERC20 tokens.

lukechilds: And there's no harm in the duplicate enabling of `alwaysEnabledCurrencies` that are also in `enabledCoins` because mm will just instantly return success saying they're already activated.

sindresorhus: I don't know how slow the `enableCoin` API is, but you're now enabling KMD and CHIPS twice, which seems moot.

sindresorhus: You also need to rebase from master now and rename `api.enableCoin` to `api.enableCurrency`

sindresorhus: Can't you just call `api.enableCoin('ETOMIC')` first then? That would make for more readable code too.

lukechilds: Did you miss my last comment?

The perf impact is negligable, it takes ~2ms to activate an already active coin.

screen shot 2018-05-29 at 6 15 30 pm

lukechilds: >Can't you just call api.enableCoin('ETOMIC') first then? That would make for more readable code too.

Yeah, I thought this was neater, and no harm double making sure KMD/CHIPS are enabled. But maybe `api.enableCoin('ETOMIC');` is better actually.

sindresorhus: We both know `mm` performance is not predictable, but I mostly think `api.enableCoin('ETOMIC')` is better as it's explicit about the intent.

PR #330, opened 2018-05-29
Refactor `api.enableCoin()`

lukechilds:
This is required for ETH/ERC20, that's not ready yet but this is still an improvement over the current code so I've cherry picked it out from my local ETH branch.

This way is much neater, it just auto detects if we have Electrum servers for a given currency symbol. If we do it's enabled in Electrum mode, if not it falls back to full node. No need to manually specify which.

We will never actually use any currencies in full node mode but ETH/ERC20 tokens need to be enabled with the full node API, even though they connect to a light client.

Also updated all terminology `coin => currency`.
Also no longer expose different methods, it's just consolidated into a single `api.enableCurrency(symbol)` method.

### DIFF
--- a/app/renderer/containers/App.js
+++ b/app/renderer/containers/App.js
@@ -173,9 +173,9 @@ class AppContainer extends Container {
 		};
 	}
 
-	enableCoin(coin) {
+	enableCurrency(coin) {
 		this.setState(prevState => {
-			this.api.enableCoin(coin);
+			this.api.enableCurrency(coin);
 			const enabledCoins = [...prevState.enabledCoins, coin];
 			config.set('enabledCoins', enabledCoins);
 			return {enabledCoins};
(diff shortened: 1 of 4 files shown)


app/renderer/containers/App.js
```diff
@@ -173,9 +173,9 @@ class AppContainer extends Container {
 		};
 	}
 
-	enableCoin(coin) {
+	enableCurrency(coin) {
```
sindresorhus: You need to rename `disableCoin` too then.

sindresorhus:
Good cleanup!

app/renderer/containers/App.js
lukechilds: I will do all `coin => currency` at some point but just did this as I was working on it.

Can we leave all the `coin => currency` till after ETH is merged? Just so I don't get a load of conflicts.

sindresorhus: Sure

PR #333, opened 2018-05-30
Fix number inputs

sindresorhus:
Basically, React is totally broken when it comes to number inputs. The fix is to just use normal string based inputs. I also added some validation for number inputs.

I also changed the withdraw modal to use separate states for the amount and the amount in USD. That's the only way to make it work.

tl;dr Numbers are hard

### DIFF
--- a/app/renderer/components/Input.js
+++ b/app/renderer/components/Input.js
@@ -2,14 +2,33 @@ import React from 'react';
 import {classNames} from 'react-extras';
 import './Input.scss';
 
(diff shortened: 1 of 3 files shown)


lukechilds:
Sorry, accidental close.

When I checkout this branch I can't log in. As soon as I enter a char for my password it sets the input value as `NaN`. Slight inconvenience 😉

sindresorhus:
> When I checkout this branch I can't log in. As soon as I enter a char for my password it sets the input value as NaN. Slight inconvenience 😉

Oops, fixed.

PR #334, opened 2018-05-30
Confirm closing the app if there are active swaps

### LINKED ISSUE #229: Confirm closing the app if there are active swaps
With a message that the trade might be ruined if you close the app. Sometimes marketmaker can kickstart a interrupted swap, sometimes not. Nobody knows.

https://github.com/lukechilds/hyperdex-bugtracker/issues/1#issuecomment-389810248

> Although there's no real way for us to know if the swap is actually swapping normally or if it's "stuck". But still good to show a warning either way.
> \- @lukechilds

sindresorhus:
Fixes #229

screen shot 2018-05-30 at 22 34 42

### DIFF
--- a/app/constants.js
+++ b/app/constants.js
@@ -28,3 +28,5 @@ exports.alwaysEnabledCurrencies = [
 	'KMD',
 	'CHIPS',
 ];
+
+exports.launchTimestamp = Date.now();
--- a/app/constants.js
+++ b/app/constants.js
@@ -29,4 +29,4 @@ exports.alwaysEnabledCurrencies = [
 	'CHIPS',
 ];
 
-exports.launchTimestamp = Date.now();
+exports.appTimeStarted = Date.now();
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -4,7 +4,7 @@ import _ from 'lodash';
 import {Container} from 'unstated';
 import appContainer from 'containers/App';
 import fireEvery from '../fire-every';
-import {launchTimestamp} from '../../constants';
+import {appTimeStarted} from '../../constants';
 
 const getInitialState = () => ({
 	baseCurrency: 'CHIPS',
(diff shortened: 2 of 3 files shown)


app/renderer/containers/Exchange.js
```diff
@@ -109,4 +111,36 @@ class ExchangeContainer extends Container {
...
+		return swap.timeStarted > launchTimestamp && !['completed', 'failed'].includes(swap.status);
+	});
+
+	if (hasInProgressSwaps) {
+		event.returnValue = true;
+
+		const selectedButtonIndex = api.dialog.showMessageBox(activeWindow(), {
+			type: 'question',
+			title: 'Are you sure you want to quit?',
+			message: 'You have swap in-progress. Blah blah.',
```
sindresorhus: @lukechilds Any suggestion for the text?

app/constants.js
```diff
@@ -28,3 +28,5 @@ exports.alwaysEnabledCurrencies = [
 	'KMD',
 	'CHIPS',
 ];
+
+exports.launchTimestamp = Date.now();
```
sindresorhus: @lukechilds Do you think this should be just a number timestamp like now or a `Date` object? I can't decide.

app/renderer/containers/Exchange.js
sindresorhus: Something like:

> You have swaps in-progress. We'll try to continue the swaps the next time you run the app, but we can't promise it will work.

app/constants.js
lukechilds: I always prefer timestamps over `Date` objects.

Also, `swap.timeStarted` is a timestamp so nice to keep them the same.

app/renderer/containers/Exchange.js
lukechilds: I think

>but we can't promise it will work.

should be

>but we recommend you leave HyperDEX running until the swap completes.

It *should* 100% always either continue the swap or unwind and reclaim the original funds after a restart.

It depends based on how much time has passed, if the other peer is still online etc. However we've yet to test this thoroughly but if this isn't how it works it's a major bug in `marketmaker` and needs fixing.

app/constants.js
lukechilds: 👌

PR #336, opened 2018-05-31
Add `SuperContainer`

sindresorhus:
This is a more powerful container that lets us listen to a connected component's lifecycle hooks, and also reset the state of the container if `getInitialState` is used.

### DIFF
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -1,32 +1,33 @@
 /* eslint-disable react/no-access-state-in-setstate */
 import _ from 'lodash';
-import {Container} from 'unstated';
+import SuperContainer from 'containers/SuperContainer';
 import appContainer from 'containers/App';
 import fireEvery from '../fire-every';
 
-const getInitialState = () => ({
-	baseCurrency: 'CHIPS',
-	quoteCurrency: 'KMD',
-	activeSwapsView: 'All',
-	swapHistory: [],
-	orderBook: {
-		bids: [],
-		asks: [],
-		biddepth: 0,
-		askdepth: 0,
-	},
-	isSendingOrder: false,
-});
+class ExchangeContainer extends SuperContainer {
+	getInitialState() {
+		return {
+			baseCurrency: 'CHIPS',
+			quoteCurrency: 'KMD',
+			activeSwapsView: 'All',
+			swapHistory: [],
+			orderBook: {
+				bids: [],
+				asks: [],
+				biddepth: 0,
+				askdepth: 0,
+			},
+			isSendingOrder: false,
+		};
+	}
 
-class ExchangeContainer extends Container {
-	state = getInitialState();
+	componentDidMountOnce() {
+		this.setSwapHistory();
+		appContainer.swapDB.on('change', this.setSwapHistory);
+	}
 
 	constructor() {
 		super();
-		this.setSwapHistory();
-		appContainer.getSwapDB.then(swapDB => {
-			swapDB.on('change', this.setSwapHistory);
-		});
 
 		appContainer.subscribe(() => {
 			if (!appContainer.state.enabledCoins.includes(this.state.baseCurrency)) {
@@ -45,8 +46,7 @@ class ExchangeContainer extends Container {
 	}
 
 	setSwapHistory = async () => {
-		const swapDB = await appContainer.getSwapDB;
-		this.setState({swapHistory: await swapDB.getSwaps()});
+		this.setState({swapHistory: await appContainer.swapDB.getSwaps()});
 	};
 
 	setBaseCurrency(baseCurrency) {
@@ -55,7 +55,7 @@ class ExchangeContainer extends Container {
 			this.setState({quoteCurrency: this.state.baseCurrency});
 		}
 
-		const {orderBook} = getInitialState();
+		const {orderBook} = this.getInitialState();
 		this.setState({baseCurrency, orderBook});
 		this.fetchOrderBook();
 	}
@@ -65,7 +65,7 @@ class ExchangeContainer extends Container {
 			this.setState({baseCurrency: this.state.quoteCurrency});
 		}
 
-		const {orderBook} = getInitialState();
+		const {orderBook} = this.getInitialState();
 		this.setState({quoteCurrency, orderBook});
 		this.fetchOrderBook();
 	}
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -21,7 +21,7 @@ class ExchangeContainer extends SuperContainer {
 		};
 	}
 
-	componentDidMountOnce() {
+	componentDidInitialMount() {
 		this.setSwapHistory();
 		appContainer.swapDB.on('change', this.setSwapHistory);
 	}
--- a/app/renderer/containers/SuperContainer.js
+++ b/app/renderer/containers/SuperContainer.js
@@ -88,21 +88,28 @@ class SuperContainer extends Container {
 
 	resetState() {
 		if (this.getInitialState) {
-			this.state = this.getInitialState();
+			const state = this.getInitialState();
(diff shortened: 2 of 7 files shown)


app/renderer/containers/Exchange.js
```diff
@@ -1,32 +1,33 @@
...
+				bids: [],
+				asks: [],
+				biddepth: 0,
+				askdepth: 0,
+			},
+			isSendingOrder: false,
+		};
+	}
+
+	componentDidMountOnce() {
```
sindresorhus: Happy to consider suggestions for a better name for this.

app/renderer/containers/SuperContainer.js
```diff
@@ -0,0 +1,113 @@
...
+	ClassComponent.displayName = FunctionalComponent.displayName || FunctionalComponent.name;
+	hoistNonReactStatics(ClassComponent, FunctionalComponent);
+
+	return ClassComponent;
+};
+
+const addLifeCycleHooks = (self, lifecycleHooks) => {
+	for (const [key, value] of Object.entries(lifecycleHooks)) {
+		if (key === 'constructor') {
+			value(self.props);
```
lukechilds: maybe `key`/`value` would be better named as `methodName`/`method` or `hookName`/`hook`.

I had to do a double take on `value(self.props)` to grok what was going on. I think something like:

```js
if (methodName === 'constructor') {
	method(self.props);
}
```

is much more clear.

sindresorhus: Yup, much clearer.

app/renderer/containers/SuperContainer.js
```diff
@@ -0,0 +1,113 @@
...
+			continue;
+		}
+
+		self[key] = value;
+	}
+};
+
+// eslint-disable-next-line no-unused-vars
+const withState = (FunctionalComponent, initialState = {}, lifecycleHooks = {}) => {
```
lukechilds: As far as I can tell, this is never used anywhere, is that intentional?

app/renderer/containers/SuperContainer.js
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	resetState() {
+		if (this.getInitialState) {
+			this.state = this.getInitialState();
+		}
+	}
+
+	// Connect the container to a component to receive some lifecycle hooks
+	connect(component) {
```
lukechilds: So this is copying `componentDidMount` and `componentWillUnmount` and your extra `once` methods from the component over to the container?

I'm kind of confused how this works, because it looks like you're returning the component from `connect()`, not the container itself.

lukechilds: Oh, wait, now I see, I was assuming:

```js
export default exchangeContainer.connect(Exchange);
```

we exporting the container. But it's not, it's exporting the component.

(viewing the diff on GH, not the individual files locally)

lukechilds: Does this mean we can't connect to a class component because `withLifecycleHooks` only accepts a functional component?

app/renderer/containers/SuperContainer.js
```diff
@@ -0,0 +1,113 @@
...
+		self.state = initialState;
+		addLifeCycleHooks(self, lifecycleHooks);
+
+		self.render = () => FunctionalComponent({
+			...self.props,
+			state: self.state,
+			setState: self.setState.bind(self),
+		});
+	});
+};
```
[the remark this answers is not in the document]
lukechilds: This is really cool.

app/renderer/containers/Exchange.js
lukechilds: I think `initalMount()` is nice but kind of breaks the React naming convention.

Possibly `componentDidInitialMount()`?

I think `initial` expresses intent better than `once`.

app/renderer/containers/SuperContainer.js
sindresorhus: > So this is converting a functional component into a class component and hooking our container lifecycle methods into the component lifecycle methods. Right?

Correct. I'll add some code comments to clarify this.

> Does this mean we can't connect to a class component because withLifecycleHooks only accepts a functional component?

I didn't think about that use-case. I'll add support for it.

sindresorhus:
It now also supports class components => https://github.com/lukechilds/hyperdex/pull/336/commits/7355268a8138a3e49545a57469c3a8a8d8cee78e

You can test by replacing `views/Exchange/index.js` with this:

```js
import React from 'react';
import {Subscribe} from 'unstated';
import exchangeContainer from 'containers/Exchange';
import TabView from 'views/TabView';
import Intro from './Intro';
import Order from './Order';
import Chart from './Chart';
import Swaps from './Swaps';
import './Exchange.scss';

class Exchange extends React.Component {
	render() {
		exchangeContainer.watchOrderBook();

		return (
			<Subscribe to={[exchangeContainer]}>
				{() => (
					<React.Fragment>
						<Intro/>
						<TabView className="Exchange">
							<Order type="buy"/>
							<Order type="sell"/>
							<Chart/>
							<Swaps/>
						</TabView>
					</React.Fragment>
				)}
			</Subscribe>
		);
	}
}

export default exchangeContainer.connect(Exchange);
```

app/renderer/containers/SuperContainer.js
lukechilds: What should happen if we try to connect to two separate components. The first one that mounts will trigger the container lifecycle hooks right.

Do you think that's ok? Or should it throw if you try to connect to an already connected container?

lukechilds:
This looks absolutely brilliant.

So just to recap, the main benefit this gives us, is that we can set things up in container lifecycle hooks rather than constructors. These won't fire until the component has mounted, and as long as the component will only be mounted after login, we can be sure all the login process has completed and therefore things like portfolio/swapDB/api are available?

And we can write all our components as functional components and they will automatically be converted to class components to hook into the lifecycle methods.

app/renderer/containers/SuperContainer.js
```diff
@@ -0,0 +1,132 @@
...
+	constructor(...args) {
+		super(...args);
+		this.resetState();
+	}
+
+	resetState() {
+		if (this.getInitialState) {
+			this.state = this.getInitialState();
+		}
+	}
```
lukechilds: Do you think it would be worth setting the state to an empty object if `this.getInitialState` is `undefined`?

Calling `resetState()` and having it do nothing doesn't seem intuitive.

lukechilds: Also, how come you directly set state as opposed to calling `this.setState()`.

I was under the impression you needed to call `this.setState()` for it to cause a re-render.

app/renderer/containers/Exchange.js
sindresorhus: > Possibly componentDidInitialMount()?

👍

sindresorhus:
> So just to recap, the main benefit this gives us, is that we can set things up in container lifecycle hooks rather than constructors. These won't fire until the component has mounted, and as long as the component will only be mounted after login, we can be sure all the login process has completed and therefore things like portfolio/swapDB/api are available?

Exactly

> And we can write all our components as functional components and they will automatically be converted to class components to hook into the lifecycle methods.

Yup

app/renderer/containers/SuperContainer.js
sindresorhus: > Do you think it would be worth setting the state to an empty object if this.getInitialState is undefined?

I think that would go against expected behavior as the component might depend on the initial state. I would rather throw then.

sindresorhus:
Everything should be addressed in https://github.com/lukechilds/hyperdex/pull/336/commits/45b3df0280b390dbc44e8bd58032be60bbf622ad and https://github.com/lukechilds/hyperdex/pull/336/commits/c853b40aa0e4aaf657e90297292e13f2ec79e6a3

lukechilds:
This is awesome man!

sindresorhus:
giphy

PR #337, opened 2018-06-01
Add Komodo Asset Chain icons

sindresorhus:
Icons created by Nadia Sotnikova: http://nadiasotnikova.com

This also refactors the app to use the `CurrencyIcon` component everywhere.


lukechilds:
Nice they fit in very nicely with the existing icons.

```
// TODO: Read this list from disk instead. Not a priority though.
```

Would be nice if we did this for all icons at some point. Currently dev tools gets spammed with errors for all the missing icons.

Not a high priority though.

PR #338, opened 2018-06-01
ETH support

lukechilds:
This is still a WIP. It's mostly done on our end but currently blocked by a few issues in `marketmaker` before we can merge:

~~https://github.com/artemii235/SuperNET/issues/56~~ Resolved
~~https://github.com/artemii235/SuperNET/issues/57~~ Resolved

These issues are related but don't need to block our work:
https://github.com/artemii235/SuperNET/issues/58
https://github.com/artemii235/SuperNET/issues/59


lukechilds:
Super nasty hack if you just wanna see your balance:

```diff
diff --git a/app/renderer/containers/App.js b/app/renderer/containers/App.js
index 2f70f6ed..9f10ad3d 100644
--- a/app/renderer/containers/App.js
+++ b/app/renderer/containers/App.js
@@ -109,6 +109,7 @@ class AppContainer extends Container {
 			this.stopWatchingCurrencies = await fireEvery(async () => {
 				const {price: kmdPriceInUsd} = this.coinPrices.find(x => x.symbol === 'KMD');
 				let {portfolio: currencies} = await this.api.portfolio();
+				const {balance: ethBalance} = await _api.balance('ETH', '0x7b0f414f040f59a743716027f8a24a37dbe1d16f');
 
 				// TODO(sindresorhus): Move the returned `mm` currency info to a sub-property and only have cleaned-up top-level properties. For example, `mm` has too many properties for just the balance.
 
@@ -117,6 +118,10 @@ class AppContainer extends Container {
 					.map(currency => {
 						currency.symbol = currency.coin; // For readability
 
+						if (currency.symbol === 'ETH') {
+							currency.balance = ethBalance;
+						}
+
 						const {cmcPriceUsd, cmcPercentChange24h} = this.getCurrencyPrice(currency.symbol);
 
 						if (cmcPriceUsd) {
```

Or you can view from the console:

```js
_api.debug({method: 'balance', coin: 'ETH', address: '0x7b0f414f040f59a743716027f8a24a37dbe1d16f'})
```

(obvs replace your address in the examples)

app/renderer/views/Exchange/Order.js
```diff
@@ -59,8 +65,8 @@ const Center = props => {
 					<thead>
 						<tr>
 							<th>Price ({state.quoteCurrency})</th>
-							<th>Avg Vol</th>
-							<th>Max Vol</th>
+							<th>{!isEtomic && 'Avg Vol'}</th>
+							<th>{!isEtomic && 'Max Vol'}</th>
```
sindresorhus:
```js
{!isEtomic &&
	<th>Avg Vol</th>
	<th>Max Vol</th>
}
```

Would be better. Fewer checks and then we don't include the additional `<th>`s either. Including empty `<th>` has no effect now, but it really depends on the CSS and could change in the future.

app/marketmaker/supported-currencies.js
```diff
@@ -394,6 +394,27 @@ const supportedCurrencies = [
 			},
 		],
 	},
+	{
+		coin: 'ETH',
+		etomic: '0x0000000000000000000000000000000000000000',
+		rpcport: 80,
+	},
```
sindresorhus: Can you add `MYTH` too? For testing.

lukechilds: I have it locally, need more testing before committing to make sure decimals display correctly. Also you can't do any trades because there are no MYTH Bobs. I plan to set one up.

lukechilds:
@sindresorhus @kevva This should be working now. I need to do a bit more testing for ERC20 swaps. They should work but some values will be wrong. I'll address that in a followup PR.

The withdraw code is kind of messy :/
I've abstracted it into a simple `withdraw({currency, address, amount})` method which will work across BTC forks and ETH/ERC20 tokens. However because the mm APIs are quite different for BTC/ETH, there's [a lot of glue code](https://github.com/lukechilds/hyperdex/blob/6f97c72919a1af65770f0b74146234eb09aa2dec/app/renderer/api.js#L190-L307) to make them both work nicely. Let me know if you've got any suggestions on how to simplify that.

app/renderer/api.js
```diff
@@ -258,6 +245,67 @@ export default class Api {
...
+		};
+
+		return {
+			txFee,
+			broadcast,
+		};
+	}
+
+	withdraw(opts) {
+		if (typeof opts.currency !== 'string') {
```
kevva: Maybe use `ow` for these checks now that #353 is merged?

app/renderer/containers/App.js
```diff
@@ -110,36 +110,38 @@ class AppContainer extends Container {
...
-					currency.cmcBalanceUsdFormatted = formatCurrency(currency.cmcBalanceUsd);
+						currency.balanceFormatted = roundTo(currency.balance, 8);
+						currency.cmcPriceUsdFormatted = formatCurrency(currency.cmcPriceUsd);
+						currency.cmcBalanceUsdFormatted = formatCurrency(currency.cmcBalanceUsd);
 
-					return currency;
-				});
+						return currency;
+					})
+					.filter(currency => !hiddenCurrencies.includes(currency.symbol));
```
kevva: Better to `filter` before to avoid unnecessary mapping.

app/renderer/api.js
kevva: Maybe use `ow` for these checks now that #353 is merged?

app/renderer/containers/App.js
kevva: Better to filter before to avoid unnecessary mapping.

lukechilds: The only reason I did it like this is because we have decided to use different terminology to the market maker api.

With the mm api `coin` means symbol. In our code we don't use the word `coin`, we have a `currency` object which contains a `symbol` property.

<sub>(we do actually have some references of `coin` leftover in our code from before we made that decision, this will be updated soon).</sub>

But I agree, normally I would filter first, this was just so I could use `currency.symbol` which only exists after the map. I can filter before the map and use `currency.coin`. Maybe I was being overly pedantic.

@sindresorhus what do you think?

app/renderer/api.js
lukechilds: Done in latest merge 👍

app/renderer/containers/App.js
sindresorhus: I would filter first too.

sindresorhus:
Yay! The day we've all been waiting for. Great work, @lukechilds 🙌

PR #343, opened 2018-06-05
Gracefully handle Electrum errors

### LINKED ISSUE #187: Handle electrum errors during login
Currently if an electrum server fails for a single currency, the entire login process fails.

We should allow them to login but notify them that the electrum servers are down for that currency.

sindresorhus:
It now shows a system notification if it fails.

Fixes #187

### DIFF
--- a/app/marketmaker/supported-currencies.js
+++ b/app/marketmaker/supported-currencies.js
@@ -554,11 +554,11 @@ const supportedCurrencies = [
 		coin: 'KMD',
 		electrumServers: [
 			{
-				host: 'force.fail',
+				host: 'electrum1.cipig.net',
 				port: 10001,
 			},
 			{
-				host: 'force.fail',
+				host: 'electrum2.cipig.net',
 				port: 10001,
 			},
 		],
--- a/app/marketmaker/supported-currencies.js
+++ b/app/marketmaker/supported-currencies.js
@@ -554,11 +554,11 @@ const supportedCurrencies = [
 		coin: 'KMD',
 		electrumServers: [
 			{
-				host: 'sindresorhus.com',
+				host: 'electrum1.cipig.net',
 				port: 10001,
 			},
 			{
-				host: 'sindresorhus.com',
+				host: 'electrum2.cipig.net',
 				port: 10001,
 			},
 		],
--- a/app/marketmaker/supported-currencies.js
+++ b/app/marketmaker/supported-currencies.js
@@ -554,11 +554,11 @@ const supportedCurrencies = [
 		coin: 'KMD',
 		electrumServers: [
 			{
-				host: 'electrum1.cipig.net',
+				host: 'force.fail',
 				port: 10001,
 			},
 			{
-				host: 'electrum2.cipig.net',
+				host: 'force.fail',
 				port: 10001,
 			},
 		],
--- a/app/marketmaker/supported-currencies.js
+++ b/app/marketmaker/supported-currencies.js
@@ -554,11 +554,11 @@ const supportedCurrencies = [
 		coin: 'KMD',
 		electrumServers: [
 			{
-				host: 'electrum1.cipig.net',
+				host: 'sindresorhus.com',
 				port: 10001,
 			},
 			{
-				host: 'electrum2.cipig.net',
+				host: 'sindresorhus.com',
 				port: 10001,
 			},
 		],
--- a/app/renderer/api.js
+++ b/app/renderer/api.js
@@ -82,7 +82,16 @@ export default class Api {
 			}));
 
 			const responses = await Promise.all(requests);
-			return responses.filter(response => response.result === 'success') > 0;
+			const success = responses.filter(response => response.result === 'success').length > 0;
+
+			if (!success) {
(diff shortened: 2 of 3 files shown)


sindresorhus:
This doesn't currently work as it seems Marketmaker blocks enabling other Electrum servers when one of them is not responding correctly.

I'm just getting this logged after 20 seconds:

screen shot 2018-06-05 at 20 33 06

Even though it's only `KMD` that has invalid Electrum servers.


@lukechilds Any thoughts on what we should/can do?

app/marketmaker/supported-currencies.js
```diff
@@ -554,11 +554,11 @@ const supportedCurrencies = [
 		coin: 'KMD',
 		electrumServers: [
 			{
-				host: 'electrum1.cipig.net',
+				host: 'sindresorhus.com',
 				port: 10001,
 			},
 			{
-				host: 'electrum2.cipig.net',
+				host: 'sindresorhus.com',
```
sindresorhus: This is just here for testing.

lukechilds:
Testing now, I may have an idea.

lukechilds:
The issue is that you're timing out after 20 seconds, but it takes way longer than that for mm to time out. And mm processes all requests serially, so the first request to fail will block all future requests meaning they all take longer than 20 seconds (but do complete successfully).

screen shot 2018-06-08 at 1 29 50 pm

Instead of implementing our own timeout, we need to check the mm response.

Failures respond with:

```js
{"ipaddr":"sindresorhus.com","port":10001,"error":"couldnt connect to electrum server"}
```

successes respond with:

```js

**Comment by lukechilds** [comment]
Just realised, this is already handled by the api:

```js
return responses.filter(response => response.result === 'success') > 0;
```

lukechilds:
Working on a fix.

sindresorhus:
> And mm processes all requests serially, so the first request to fail will block all future requests meaning they all take longer than 20 seconds (but do complete successfully).

I had a feeling it would be something like that.

> but it takes way longer than that for mm to time out.

How long? We don't want it to be too long. That just creates a bad user-experience if the login takes forever just because of one non-working Electrum server.

lukechilds:
>How long?

It looks like not resolving to DNS takes about 5ms (my `force.fail` example).
However if I use a real domain that just isn't an Electrum server it seems to consistently take 1.3 mins 😬

>That just creates a bad user-experience if the login takes forever just because of one non-working Electrum server.

Agreed.

@jl777 is the 1.3 min timeout hardcoded into `marketmaker`? If so is it possible to reduce it? Realistically, if an Electrum server is running slow, we probably don't want to connect to it anyway. It might reduce the chance of successful swaps. I think something like 5 seconds is more than enough.

sindresorhus:
> I think something like 5 seconds is more than enough.

5 seconds might too short. Especially in countries with generally slower networks. I think 10 seconds would be good.

jl777:
electrum timeout is set to a relatively low value. 7 or 13 seconds. I think it is just getting stuck returning from a tcp connect. Not sure why it is taking over one minute...

jl777:
does this happen if you use IP address? if so, then the issue is DNS

lukechilds:
>electrum timeout is set to a relatively low value. 7 or 13 seconds. I think it is just getting stuck returning from a tcp connect. Not sure why it is taking over one minute...

@jl777 would it be possible to just implement a timer in `marketmaker`? If a connection hasn't been established successfully after the time has elapsed for whatever reason just abort everything and return an error.

>does this happen if you use IP address? if so, then the issue is DNS

Yeah, just tried with 216.58.199.206 (Google's IP) and it took 1.3 minutes

jl777:
how to abort a pending connect() call? maybe there is a way, but I am not sure how to do it in any OS independent way

jl777:
reproduced the issue, I will see what I can do

jl777:
all the time is in the connect() system call

        printf("call connect sock.%d\n",sock);
        result = connect(sock,(struct sockaddr *)&saddr,addrlen);
        printf("called connect result.%d lag.%d\n",result,(int32_t)(time(NULL) - starttime));

I have an idea to fix this programmatically, but no idea if it will work or not

sindresorhus:
> how to abort a pending connect() call?

If there's no support for it in the system, you can just abstract the call, set a timer, when the timer runs out, fail the method, and when the response comes back from `connect()`, just ignore it. That's how we do it in the JS world with Promises that cannot be canceled.

jl777:
the issue is that mm is single threaded for processing and what you say is easy, requires implementing a totally async calling structure, which wont be until mm2.0

i am experimenting with changing socket from sync to async to sync

lukechilds:
Thanks James

jl777:
pushed a version that times out in 5 seconds for the invalid electrum ipaddr above, it seems to still work ok in unix, we are testing in windows which of course behaves differently

lukechilds:
@jl777 This is working for me now (timing out after 5s)

@sindresorhus I think this is ok to merge, the timeout issue is unrelated to HyperDEX and even without this PR it will still delay for ages, they just won't get any feedback.

PR #344, opened 2018-06-06
Refactor the list of swaps into a separate component

sindresorhus:


sindresorhus:
> I would really like to show the cancel button in the exchange view. Not important for this PR, but I think it's important to show it in the exchange view.

I think we need to make the Cancel button actually work before we put it in the Exchange view. Right now it's more annoying than it's useful because of marketmaker limitations...

lukechilds:
Yeah, I think the positioning is exacerbating the issue though. It does work reliably if you click it before the trade gets matched, which is a pretty short time frame so only really possible from the exchange view.

But I agree, it's very confusing at the moment, I have some ideas on how to make it work a little better.

sindresorhus:
But for active currencies like KMD/DOGE it gets matched almost right away, so I wouldn't be able to click `Cancel`. I honestly don't think cancel is useful until we make it also work *after* it gets matched.

lukechilds:
I did ask James previously to allow cancelling in-progress swaps but he was strongly opposed to it.

I will ask him again.

lukechilds:
@sindresorhus https://github.com/lukechilds/hyperdex/issues/262#issuecomment-395390245

PR #346, opened 2018-06-07
Avoid multiple state mutations when calculating amount and total

### LINKED ISSUE #268: Small UI bug when setting up price/amount/total in the Exchange view
OS: Ubuntu 18.04 with HyperDex .deb installer
When setting up a trade, the Total doesn't update if I first set the Amount and then the Price

---

HyperDEX 0.1.0-alpha.1
Electron 2.0.0
linux 4.15.0-20-generic
Locale: it

kevva:
This should fix side effects that can happen when calling `setState` multiple times calculating the total and amount when making an order. Given that `setState` is async, it can result in the wrong values being used.

Fixes #268.

kapture 2018-06-08 at 0 54 20

### DIFF
--- a/app/renderer/views/Exchange/Order.js
+++ b/app/renderer/views/Exchange/Order.js
@@ -270,16 +270,16 @@ class Order extends React.Component {
 	handlePriceChange = price => {
 		price = String(price);
 
-		this.setState(prevState => ({
-			price,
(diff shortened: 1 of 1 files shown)


sindresorhus:
Perfect. 👌

unicorn18

PR #347, opened 2018-06-08
Split Settings into Portfolio and App Settings

sindresorhus:
Since some settings are only available when logged out and some are only available when logged in.

Clicking the "Log Out" links logs the user out and takes them to the app settings.

The "Theme" setting (that I'll be working on next), for example, will be available on both.

They are two different components for simplicity, since the portfolio setting component might depend on logged-in state, but they share most of the CSS.

screen shot 2018-06-08 at 16 50 10
screen shot 2018-06-08 at 16 50 19


sindresorhus:
@lukechilds I'm not happy about the back button in the app settings, but I'm not sure where else to place it. Maybe we could place on the left side of the "Settings" header, like on iOS, without the background and border.

sindresorhus:
When logged out, you can only access the settings from the menu bar menu "HyperDEX" => "Preferences…", I guess you want to make it more visible by placin it somewhere in the Login UI? Maybe as a link below "Forgot Password".

sindresorhus:
> Could we show all the App settings inside the portfolios settings, but not the other way round?

But part of the reason for splitting them is that many of the app settings can't be set while logged in, like "Marketmaker URL". We can add the ones that can be set, like the upcoming "Theme" setting.

lukechilds:
Yeah, the mm url should be the only one that requires a restart. And we could always just add `* requires a restart` to it and auto restart after save.

sindresorhus:
> Yeah, the mm url should be the only one that requires a restart.

You previously mentioned custom Electrum servers too, which will also require logging in again.

> And we could always just add * requires a restart to it and auto restart after save.

We don't have a save button. Settings are just saved live. We could show a link to restart, but it's just easier to go to the app dedicated settings, I think.

lukechilds:
Good point 👍

app/renderer/views/AppSettings.js
```diff
@@ -0,0 +1,69 @@
+import {remote} from 'electron';
+import React from 'react';
+import _ from 'lodash';
```
kevva: Not that it matters in an `electron` app, but `import debounce from 'lodash/debounce'` would make the bundle smaller.

lukechilds: Won't Webpack do tree shaking and handle this for us? Not sure it will, genuine question.

sindresorhus:
This should be done now.

I needed to be able to dynamically style the button, so i decided to add `styled-jsx`. I know we early on decided not to use CSS-in-JS, but there's really no other way to be able to dynamically specify a hover style. `styled-jsx` is very lightweight and we can use it only when we need the dynamic behavior. I've also made a `<Svg>` component which makes it so much easier to use and modify icons. Previously I hacked icon styling with `-webkit-mask` CSS, but it's limiting.

app/renderer/views/AppSettings.js
sindresorhus: Webpack still doesn't really do tree shaking very well, even if we did `import debounce from 'lodash/debounce'`. Doesn't matter for us though.

sindresorhus:
> I think we could probably remove the "Portfolio" section form the app settings. It makes sense you have to login to view them, I don't think we need to point it out.

Yeah, I agree, it's over the top. Fixed.

PR #350, opened 2018-06-11
Consolidate currencies that ignore external price

lukechilds:
Probably better to have these in a constant than in different files.

Also, if we're ignoring CMC fiat price we'll likely also wan't to ignore the fiat history and vice versa, so makes sense to consolidate this into a single set.

### DIFF
--- a/app/renderer/containers/Dashboard.js
+++ b/app/renderer/containers/Dashboard.js
@@ -8,6 +8,8 @@ import {formatCurrency} from '../util';
 import {ignoreExternalPrice} from '../../constants';
 import fireEvery from '../fire-every';
 
+const noPriceHistory = new Set();
+
 class DashboardContainer extends Container {
 	state = {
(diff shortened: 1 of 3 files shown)


app/renderer/containers/Dashboard.js
```diff
@@ -176,7 +169,7 @@ class DashboardContainer extends Container {
 		}
 
 		if (!json || json.Data.length === 0) {
-			noPriceHistory.add(symbol);
+			ignoreExternalPrice.add(symbol);
```
lukechilds: I've left this in from previous code, but seems kind of strange having a set in a file called constants that we add new values to.

Maybe this should be a separate empty `noPriceHistory` set that we add blank responses to.

Thoughts?

sindresorhus: > Maybe this should be a separate empty noPriceHistory set that we add blank responses to.

Yes

Issue #178, opened 2018-05-07
Localization


sindresorhus:
The app should be localized into some popular languages to be more user-friendly to those users. We have not yet decided which languages.

sindresorhus:
So the plan is to localize the app into these languages:

- Chinese
- Japanese
- Korean
- Russian
- German
- French
- Spanish
- Arabic
- Turkish

PR #354, opened 2018-06-11
Add basic support for localization

kevva:
This is just a simple example of a potential solution to #178.

I went with [`i18next`](https://www.i18next.com/) since it's pretty powerful and doesn't require a lot of manual handling (like `react-intl`) while still being very extendable. It also has some [components for React](https://react.i18next.com/).

I debated with myself whether to use their HOC or their render prop component, but I thought using the HOC would be cleaner, and you can also access `t` inside component methods without having to pass it around. The problem is that it'll break all the places were we're using `setActiveView` causing the components to never render (which took me a while to figure out why lol) because we're using hard coded component names rather than importing them and using their `.name` property.

We can skip the HOC altogether if we wanted though and just `remote.require` it from the `main` process.

### DIFF
--- a/app/locale.js
+++ b/app/locale.js
@@ -19,4 +19,4 @@ i18next
 	});
 
 exports.i18n = i18next;
-exports.t = i18next.t.bind(i18next);
+exports.translate = namespaces => namespaces ? i18next.getFixedT(null, namespaces) : i18next.t.bind(i18next);
--- a/app/locale.js
+++ b/app/locale.js
@@ -1,0 +1,22 @@
+'use strict';
+const path = require('path');
+const i18next = require('i18next');
+const i18nextBackend = require('i18next-sync-fs-backend');
+const {isDevelopment} = require('./util-common');
+
+i18next
+	.use(i18nextBackend)
+	.init({
+		debug: isDevelopment,
+		lng: 'en',
+		whitelist: ['en'],
+		ns: ['login', 'menu'],
+		initImmediate: false,
+		backend: {
+			loadPath: path.join(__dirname, 'locales', '{{lng}}', '{{ns}}.json'),
+			jsonIndent: 4,
+		},
+	});
+
+exports.i18n = i18next;
+exports.t = i18next.t.bind(i18next);
--- a/app/menu.js
+++ b/app/menu.js
@@ -6,10 +6,11 @@ const config = require('./config');
 const {openGitHubIssue} = require('./util');
 const {websiteUrl, repoUrl, appViews} = require('./constants');
 const {isDevelopment} = require('./util-common');
-const {t} = require('./locale');
+const {translate} = require('./locale');
 
 const {app, BrowserWindow, shell, clipboard, ipcMain: ipc, Menu} = electron;
 const appName = app.getName();
+const t = translate('menu');
 
 const sendAction = (action, data) => {
 	const [win] = BrowserWindow.getAllWindows();
@@ -28,19 +29,19 @@ const setActiveView = view => {
 const createHelpMenu = () => {
 	const helpSubmenu = [
 		{
-			label: t('menu:help.website'),
+			label: t('help.website'),
 			click() {
 				shell.openExternal(websiteUrl);
 			},
 		},
 		{
-			label: t('menu:help.sourceCode'),
+			label: t('help.sourceCode'),
 			click() {
 				shell.openExternal(repoUrl);
 			},
 		},
 		{
-			label: t('menu:help.reportIssue'),
+			label: t('help.reportIssue'),
 			click() {
 				openGitHubIssue('<!-- Please succinctly describe your issue and steps to reproduce it -->');
 			},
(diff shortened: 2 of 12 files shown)


app/locale.js
```diff
@@ -0,0 +1,22 @@
...
+const path = require('path');
+const i18next = require('i18next');
+const i18nextBackend = require('i18next-sync-fs-backend');
+const {isDevelopment} = require('./util-common');
+
+i18next
+	.use(i18nextBackend)
+	.init({
+		debug: isDevelopment,
+		lng: 'en',
```
kevva: Hardcoded for now, but maybe we should use `app.getLocale()` and fallback to `en` instead?

sindresorhus: Yeah, just use https://github.com/neruchev/i18next-electron-language-detector

sindresorhus:
> The problem is that it'll break all the places were we're using setActiveView causing the components to never render (which took me a while to figure out why lol) because we're using hard coded component names rather than importing them and using their .name property.

I agree using that practice is better, but it *is* a bug with `i18next-react`, as HOCs should inherit the name of the wrapped component. Like done here: https://github.com/lukechilds/hyperdex/blob/703fe646221b8d3d84d84116673ba405941d0c82/app/renderer/containers/SuperContainer.js#L10-L12

sindresorhus:
> We can skip the HOC altogether if we wanted though and just remote.require it from the main process.

Wouldn't that mean we call over IPC to the main process for every `t()` call? That sounds very slow.

sindresorhus:
Any way we can have `t` available everywhere without having to wrap every single component in a HOC?

app/renderer/index.js
```diff
@@ -17,6 +21,8 @@ document.documentElement.classList.add(`os-${process.platform}`);
 
 render((
 	<Provider>
-		<App/>
+		<I18nextProvider i18n={i18n}>
+			<App/>
+		</I18nextProvider>,
```
sindresorhus: Accidental comma here

kevva:
> Any way we can have t available everywhere without having to wrap every single component in a HOC?

Yeah, that's what I meant by requiring it from the main process. But I don't know how we'd make it globally available without using context or some other hack. We could create our own HOC though with a correctly inherited component name though using https://react.i18next.com/components/i18n-render-prop, which `react-i18next` does internally anyway.

sindresorhus:
No way to precompile it, so we can just do `import {t} from './locale/menu';` or `import {t} from './locale/login';`?

kevva:
Decided to skip `react-i18next` altogether and create a wrapper around `t` that we can import on the client instead.

Are we good with this? If so, I'll start translating the rest of the texts.

app/renderer/translate.js
```diff
@@ -0,0 +1,10 @@
+import {remote} from 'electron';
+
+let instance;
+
+if (!instance) {
+	instance = remote.require('./locale').i18n;
+}
```
sindresorhus: Modules are only evaluated once, so I don't see the point in this if-statement.

sindresorhus: I am worried that every call `i18n` below will go over IPC though...

app/menu.js
```diff
@@ -27,19 +28,19 @@ const setActiveView = view => {
 const createHelpMenu = () => {
 	const helpSubmenu = [
 		{
-			label: `Website`,
+			label: t('menu:help.website'),
```
sindresorhus: Can we have a wrapper here too, so we can just do `const t = translate('menu')` at the top and then use the shorthand?

app/renderer/translate.js
lukechilds: I think you might be right re every call going over ipc. Maybe we could just have a Node.js module that manually requires the JSON files (as opposed to using `i18next-sync-fs-backend`) and exports a full `i18next` config.

Then we can require that module in node land and also `remote.require` it once in the module and have two separate instances of `i18next` but only one config file.

app/renderer/translate.js
```diff
@@ -0,0 +1,10 @@
+import {remote} from 'electron';
```
lukechilds: Much prefer the `app/renderer/translate.js` approach over wrapping everything in a HOC.

lukechilds:
>But I don't know how we'd make it globally available without using context or some other hack.

We have unstated containers (`AppContainer`) which we can use for globally accessible data.

>Decided to skip react-i18next altogether and create a wrapper around t that we can import on the client instead.
>
>Are we good with this? If so, I'll start translating the rest of the texts.

Yeah, I prefer this approach.

app/renderer/translate.js
kevva: Alright, so the translations are already available on the instance when we load it (`i18n.store.data`). We can create an instance of `i18n` on the client with that data and wrap it like we do now.

app/menu.js
kevva: Yup!

sindresorhus:
Merging this to reduce chances of merge conflicts. Do a follow-up PR with the rest.

sindresorhus:
@kevva And don't forget: https://github.com/lukechilds/hyperdex/pull/354#discussion_r194579207

kevva:
Yeah, `i18next` uses `dev` as `fallbackLng` by default. We should set it to `en`.

PR #357, opened 2018-06-13
Add some stats to the Trades view

sindresorhus:
Per discussion in Slack.

screen shot 2018-06-13 at 17 46 30

### DIFF
--- a/app/renderer/views/Trades.js
+++ b/app/renderer/views/Trades.js
@@ -87,7 +87,7 @@ export default withState(Trades, {}, {
 		exchangeContainer.setSwapHistory();
 
 		this.setState({
-			stats: await appContainer.swapDB.lastMonthStats(),
+			stats: await appContainer.swapDB.statsSinceLastMonth(),
 		});
 	},
 });
--- a/app/renderer/views/Trades.js
+++ b/app/renderer/views/Trades.js
@@ -1,10 +1,13 @@
 import React from 'react';
 import {classNames} from 'react-extras';
 import {Subscribe} from 'unstated';
+import {withState} from 'containers/SuperContainer';
+import appContainer from 'containers/App';
 import exchangeContainer from 'containers/Exchange'; // TODO(sindresorhus): Find a better place to have the SwapDB data, since both the Exchange and Trades view uses it
 import tradesContainer from 'containers/Trades';
 import View from 'components/View';
 import SwapList from 'components/SwapList';
+import {formatCurrency} from '../util';
 import AppTabView from './TabView';
 import './Exchange/Swaps.scss';
 import './Trades.scss';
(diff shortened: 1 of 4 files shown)


app/renderer/views/Trades.js
```diff
@@ -43,29 +46,48 @@ const SwapHistory = () => {
...
+export default withState(Trades, {}, {
+	async componentDidMount() {
+		/// TODO: This is only here temporarily until we move the swap stuff to the App container
+		exchangeContainer.setSwapHistory();
+
+		this.setState({
+			stats: await appContainer.swapDB.lastMonthStats(),
+		});
+	},
+});
```
sindresorhus: @lukechilds 😎

lukechilds: Niceeee, btw there is an index on `swap.timeStarted` so you should be able to query directly on ``

```js

Comment (lukechilds) [reply app/renderer/views/Trades.js]: Also might be nice to abstract it into `swapDB.statsSince(timestamp)` so we could easily reuse it for all time stats, weekly stats etc.

Comment (sindresorhus) [reply app/renderer/views/Trades.js]: Good idea. Fixed. I made a `since` and sort` option for convenience, but also `query` option that trump those for maximum power!

File: app/renderer/views/Trades.js [hunk 2]
```diff
@@ -43,29 +46,48 @@ const SwapHistory = () => {
...
+				</AppTabView>
+			);
+		}}
 	</Subscribe>
 );
 
-export default Trades;
+export default withState(Trades, {}, {
+	async componentDidMount() {
+		/// TODO: This is only here temporarily until we move the swap stuff to the App container
```
lukechilds: What do you mean by this comment?

lukechilds:
Perfect 👌

app/renderer/views/Trades.js
lukechilds: Awesome!

app/renderer/views/Trades.js
sindresorhus: I plan to move the swap state to state in the AppContainer instead of being in the Exchange container as it's also needed by the Trades view, and it's awkward import the Exchange container there.

lukechilds:
@sindresorhus I just noticed two issues with this.

1. The number of currencies should take into account unique currencies in base and quote. Currently just taking into account quote means if I do trades on `BTC/KMD`, `LTC/KMD` and `ETH/KMD` it will only display as trading 1 currency. It should be 4.

2. For whatever reason, PouchDB is actually not that fast. `swapDB.getSwaps()` takes around 250ms for me! But we have all swaps stored in `exchangeContainer` state. And these are guaranteed to be up to date because it subscribes to change events from Pouch and updates itself. So it would probably be considerably faster to just move the stats aggregation logic into `exchangeContainer` and run it over the full swap collection.

sindresorhus:
> The number of currencies should take into account unique currencies in base and quote. Currently just taking into account quote means if I do trades on BTC/KMD, LTC/KMD and ETH/KMD it will only display as trading 1 currency. It should be 4.

https://github.com/lukechilds/hyperdex/pull/362

> So it would probably be considerably faster to just move the stats aggregation logic into exchangeContainer and run it over the full swap collection.

You're basing this on just a few swaps though. What happens when the user has thousands of swaps? Eventually, we'll also want to move to pagination. So I don't think we should move away from the query interface just because it's a little bit slower on an insignificant amount of data.

lukechilds:
Fair point.

PR #358, opened 2018-06-13
Correctly handle inverse values for sell orders

lukechilds:
Additional fix for: https://github.com/lukechilds/hyperdex/pull/192

This correctly handles inverse values for sell orders. When we place a sell order `marketmaker` just inverts the values and places a buy order on the opposite pair. 

We need to account for this otherwise if we refer to the base/quote currency value in our request it will be the opposite currency in the pair to what we get in `marketmaker` responses.

Before:

<sup>Notice the values in the "Requested" section doesn't match the other sections</sup>

screen shot 2018-06-13 at 7 04 16 pm

After:

screen shot 2018-06-13 at 7 03 43 pm

### DIFF
--- a/app/renderer/components/SwapDetails.js
+++ b/app/renderer/components/SwapDetails.js
@@ -19,6 +19,27 @@ const stageToTitle = new Map([
 	['alicespend', 'Alice Spend'],
 ]);
 
+const getOverview = swap => {
+	const isBuyOrder = swap.orderType === 'buy';
+	const overview = {
+		fromTitle: 'Exchanging:',
+		forTitle: 'For:',
+		fromCurrency: isBuyOrder ? swap.quoteCurrency : swap.baseCurrency,
+		forCurrency: isBuyOrder ? swap.baseCurrency : swap.quoteCurrency,
+		fromAmount: isBuyOrder ? swap.broadcast.quoteCurrencyAmount : swap.broadcast.baseCurrencyAmount,
+		forAmount: isBuyOrder ? swap.broadcast.baseCurrencyAmount : swap.broadcast.quoteCurrencyAmount,
+	};
+
+	if (swap.executed.quoteCurrencyAmount) {
+		overview.fromTitle = 'You exchanged:';
+		overview.forTitle = 'You received:';
+		overview.fromAmount = isBuyOrder ? swap.executed.quoteCurrencyAmount : swap.executed.baseCurrencyAmount;
+		overview.forAmount = isBuyOrder ? swap.executed.baseCurrencyAmount : swap.executed.quoteCurrencyAmount;
+	}
+
+	return overview;
+};
+
 class SwapDetails extends React.Component {
 	state = {
 		isOpen: false,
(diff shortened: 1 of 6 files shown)


lukechilds:
This is easier for the users to follow, because all the values are using the same trading pair. However it is strange that the "requested" values aren't actually the values the user requested, but the values mm interpreted them as.

Alternatively we could show the real values for requested, and change the text from "sell" to "buy". That would be more accurate as to what actually happened.

Both methods of displaying the data are confusing though because the functionality is confusing. I think this is probably the best way to display it. Thoughts?

sindresorhus:
> Alternatively we could show the real values for requested, and change the text from "sell" to "buy". That would be more accurate as to what actually happened.

I would do it on all of them, `requested`, `broadcast`, `executed`. The fact that marketmaker just inverts is an implementation detail we should not expose to the user.

lukechilds:
This was a discussion I had before:

>We obviously want to hide as much of the complexity as possible from the user, but completely faking the order book value just to make it seem more intuitive is too far IMO. As far as marketmaker is concerned KMD/REVS orders are not REVS/KMD orders.

https://github.com/lukechilds/hyperdex-bugtracker/issues/15#issuecomment-387016006

That would definitely be clearer for the user, I'm not denying that, but it would mean showing completely false data. As far as mm is concerned a KMD/REVS order is not a REVS/KMD order.

It would also mean completely re-doing the swap lists and showing some REVS/KMD buys as KMD/REVS sells.

lukechilds:
In fact, yeah, I'm in favour of this, it's clearly too confusing no matter how we show it. We need to abstract it, even if it's not technically what actually happened.

Also it's useful to limit to `REVS/KMD` and get a list of the buys and sells you've made on that pair. (I wanted to do this before but it's currently impossible). The only confusing thing wit that is we should probably strip out buys and sells you made on `KMD/REVS` or the feature becomes confusing again. So you'd need to check both pairs and filter to them to view all trades youve done relating to REVS<->KMD.

sindresorhus:
Great, let's do that then 👌

lukechilds:
Ok, after these latest changes we are completely abstracting the appearance of sell orders. So if you place a sell order on `CHIPS/KMD`, `marketmaker` will invert the values and place a buy order on `KMD/CHIPS`. However we know the order was originally placed as a `CHIPS/KMD` sell so we display the order as a `CHIPS/KMD` sell and un-invert all the `KMD/CHIPS` buy responses we get from `marketmaker`.

So behind the scenes everything is just buys, but as far as the user is concerned they can place a buy or a sell and it will appear differently.

In the swap modal it will be clearly listed as a sell order and they will view the calculated inverted prices to easily track their ask price and the match price.

In the swap list the values will always be listed in the order of `baseCurrency` and then `quoteCurrency`, but if it was a sell order the red/green highlights will be switched round to easily differentiate.

`CHIPS/KMD` Buy order:

screen shot 2018-06-14 at 10 01 51 pm

`CHIPS/KMD` Sell order:

screen shot 2018-06-14 at 10 01 41 pm

Swap list:

screen shot 2018-06-14 at 10 02 33 pm

I've also removed the functionality to show both sides of a pair when you filter. This was just a hacky workaround to view buys and sells. But we don't need that any more because you can just filter to `CHIPS/KMD` and you will see both buys and sells.

jansako:
This is looking great, the trade details show what a user is expecting to see, on the swap list it is really not relevant if the order was a sell or buy since you can see which 'coin' inventory increased or decreased.

Nice progress!

lukechilds:
@sindresorhus @kevva This should be good to merge now.

Issue #175, opened 2018-05-07
Tooltips


sindresorhus:
We need a tooltip solution that use Portals and doesn't require any state handling or lots of boilerplate. It should be easy to wrap existing content with a tooltip.

sindresorhus:
Places we want tooltips:

- [ ] The "Create New Portfolio flow, at the "Seed Phrase for Your Portfolio, when the user click copy button. See the animation here: https://framer.cloud/jTdrj <img width="216" alt="screen shot 2018-06-13 at 00 35 17" src="https://user-images.githubusercontent.com/170270/41306867-ae10b42e-6ea1-11e8-84e3-97f11b509a4e.png"> 
- [ ] In the "Dashboard", select a currency, click "Deposit", and when you click copy button it should have the same tooltip and animation as the above item. See: https://framer.cloud/tKpjW
- [ ] Better tooltips for the charts. The current ones are ugly and follow the mouse. The tooltip should be pinned to their origin points.

sindresorhus:
@lukechilds @kevva I think we should also have a tooltip for the "regenerate seed phrase" button. Thoughts?
<img width="227" alt="screen shot 2018-06-25 at 11 33 16" src="https://user-images.githubusercontent.com/170270/41830487-94b668f6-786b-11e8-8f41-30ef22d317c3.png">

lukechilds:
I think most people will know what it does but definitely no harm in adding a tooltip for more clarity.

PR #360, opened 2018-06-14
Add `Tooltip` component

kevva:
* [x] Add margin to the tooltip
* [x] Add styles to the arrow for other positions than `top`
* [x] Fine tune styles (maybe some border and shadow)
* [x] Add animation while fading in (from top and down)

kapture 2018-06-14 at 15 29 59

For #175


app/renderer/components/Tooltip.js
```diff
@@ -0,0 +1,77 @@
...
+						</div>
+					)}
+				</Reference>
+				{ReactDOM.createPortal(tooltip, document.body)}
+			</Manager>
+		);
+	}
+}
+
+export default Tooltip;
```
sindresorhus: Can you add prop-types too? I've been too lazy to do this on the other components, but in hindsight, I should have.

app/renderer/components/Tooltip.scss
```diff
@@ -0,0 +1,40 @@
...
+	padding: 8px 10px;
+	transition: opacity ease-in-out;
+
+	&--entered {
+		opacity: 1;
+	}
+
+	&__arrow {
+		position: absolute;
+		width: 16px;
```
kevva: Will make these variables.

app/renderer/components/Tooltip.scss
```diff
@@ -0,0 +1,40 @@
...
+		height: 24px;
```
kevva: Will make these variables.

app/renderer/components/Tooltip.js
kevva: Yeah, I didn't because no other components had it. Will add!

app/renderer/components/Tooltip.js
```diff
@@ -0,0 +1,91 @@
...
+					</div>
+				)}
+			</Popper>
+		);
+
+		return (
+			<Manager>
+				<Reference>
+					{({ref}) => (
+						<div
```
kevva: Really bothers me that we have to wrap the children in a `div` since it won't wrap elements with absolute position correctly. I'd really like to be able to pass a `ref` to children directly. `React.forwardRef()` only works for components.

kevva:
@sindresorhus, tried to, but didn't come up with any nice border or shadows. Does the animation look good though?

kapture 2018-06-21 at 2 32 10

sindresorhus:
> tried to, but didn't come up with any nice border or shadows.

I think it needs something so it doesn't bleed into the background. Try with a slight shadow.

> Does the animation look good though?

The animation looks good.

> Also, should the content animation be in this component or should we pass another component that animates the text as content prop? Depends on how magical we want to make the tooltip.

What you have now is good. We don't need that kind of manual control.

sindresorhus:
@kevva I noticed that if I hover over the copy button, the click, then take my mouse somewhere else and then hover again, it still shows `Copied!`. I think it should revert back to the initial value when it's being hidden.

app/renderer/views/CreatePortfolio/Step2.js
```diff
@@ -5,43 +5,67 @@ import ReloadButton from 'components/ReloadButton';
...
+import Tooltip from 'components/Tooltip';
 import container from 'containers/CreatePortfolio';
 import './CreatePortfolio.scss';
 
-const CreatePortfolioStep2 = () => {
-	// TODO(sindresorhus): Fill in the link to security best practices
+class CreatePortfolioStep2 extends React.Component {
+	state = {
+		isCopied: false,
+	}
```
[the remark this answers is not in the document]
kevva: Sweet. Didn't notice that!

app/renderer/components/Tooltip.js
kevva: That's kinda cool. Doesn't seem to work though. See the screenshot below (don't mind the positioning of the elements inside the seed phrase box, it's just for testing):

screen shot 2018-06-21 at 16 28 56

kevva:
Added a slight shadow to it:

screen shot 2018-06-22 at 11 16 05

kevva:
> @kevva I noticed that if I hover over the copy button, the click, then take my mouse somewhere else and then hover again, it still shows Copied!. I think it should revert back to the initial value when it's being hidden.

Should be fixed. Added an `onClose` handler to it to reset the state.

app/renderer/components/Tooltip.js
```diff
@@ -0,0 +1,114 @@
...
+										`Tooltip--${placement}`,
+									)}
+								>
+									<div className="Tooltip__content">
+										{content}
+									</div>
+									<div
+										ref={arrowProps.ref}
+										className="Tooltip__arrow"
+										style={arrowProps.style}
```
kevva: Would've wanted to put these into `styled-jsx` too, but that'd require converting them to a CSS string. Decided not to bother with it.

app/renderer/components/Tooltip.js
sindresorhus: Ah, it's only available in Chrome 65, but Electron 2 is at 61. We'll be able to use this when Electron 3 is stable. https://github.com/hyperdexapp/hyperdex/issues/374

kevva: I added the flag to `blinkFeatures`, but it didn't work regardless (which the screenshot shows).

sindresorhus:
This looks and works perfectly now. Nice work Kevin :)

PR #362, opened 2018-06-15
Fix stats calculation for currency count

sindresorhus:
> In the last month you did 1 successful trade in 2 currencies worth $0.1139 in total

@lukechilds Is this what you had in mind?


lukechilds:
Yep, perfect 👌

PR #363, opened 2018-06-15
Stuck pending swaps

### LINKED ISSUE #279: Swap status stuck on 'Pending' for hours
Moved from: https://github.com/lukechilds/hyperdex-bugtracker/issues/10

---

> I had SN/KMD swap that stayed on 'pending' for a few hours. It stayed as 'pending' after HyperDEX restart as well.
 Not sure if it qualifies as a 'stuck' trade, since the trade never started. In any case, pending trades should probably after some time be automatically changed to 'cancelled' or whatever you decide the status for 'never started' swaps will be.
>
> image
> \- @jansako

lukechilds:
Resolves #279 

There is a bug in `marketmaker` when a swap get's some communication from a peer (like a `reserved` message) but never matches. In this scenario an error message is never sent so the swap is displayed as pending in the GUI forever.

This is a bit of a hack but it resolves the issue and should be fine until `marketmaker` v2.

It just displays swaps that are pending for more the 5 minutes as unmatched and emits a fake DB change event every minute to force a recalculation.

`marketmaker` issue: https://github.com/jl777/SuperNET/issues/775

### DIFF
--- a/app/renderer/swap-db.js
+++ b/app/renderer/swap-db.js
@@ -33,6 +33,11 @@ class SwapDB {
 		this.ready = (async () => {
 			await this.db.createIndex({index: {fields: ['timeStarted', 'uuid']}});
 			await this.migrate();
+
+			// We need to regularly check if pending swaps have timed out.
+			// https://github.com/jl777/SuperNET/issues/775
+			const ONE_MINUTE = 60000;
+			setInterval(() => ee.emit('change'), ONE_MINUTE);
 		})();
 
 		this.pQueue = new PQueue({concurrency: 1});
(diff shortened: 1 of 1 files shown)


app/renderer/swap-db.js
```diff
@@ -33,6 +33,11 @@ class SwapDB {
 		this.ready = (async () => {
 			await this.db.createIndex({index: {fields: ['timeStarted', 'uuid']}});
 			await this.migrate();
+
+			// We need to regularly check if pending swaps have timed out.
+			// https://github.com/jl777/SuperNET/issues/775
+			const ONE_MINUTE = 60000;
```
sindresorhus: `60000` => `1000 * 60` is more readable. To ensure the value is actually correct.

sindresorhus:
Yay! My trades view is finally cleaned up :D

lukechilds:
🎉 

This is also a prerequisite for a potential fix for the stuck swaps halfway through I want to try out.

PR #364, opened 2018-06-15
Disable Node.js integration for better security

### LINKED ISSUE #221: Consider isolating Node.js from the render process
Currently an XSS attack could get access to the main system due to the render process having direct access to the Node.js process.

This is low priority, an XSS attack could still steal the users balances, there's no way to protect against that (apart from taking measures against XXS in the first place which we already have).

So this would just make an absolutely terrible situation slightly less terrible.

However it would limit the attack to within HyperDEX rather than the entire system, so it's worth considering.

We already have a strict CSP and all external input is escaped by React so we are already pretty well protected against an XSS attack.

Original discussion in: https://github.com/lukechilds/hyperdex/issues/10

sindresorhus:
Fixes #221

### DIFF
--- a/app/preload.js
+++ b/app/preload.js
@@ -1,0 +1,19 @@
+'use strict';
+const {remote} = require('electron');
+
+process.once('loaded', () => {
+	global.global = global;
+	global.require = require;
+	global.process = process;
+	global.Buffer = Buffer;
+});
+
+const {require: r} = remote;
+
+global.mainModules = {
+	getPort: r('get-port'),
+	config: r('./config'),
+	portfolioUtil: r('./portfolio-util'),
+	i18n: r('./locale').i18n,
+	util: r('./util'),
+};
--- a/app/preload.js
+++ b/app/preload.js
@@ -1,19 +1,29 @@
 'use strict';
 const {remote} = require('electron');
 
+// Mock for electron-util
+// TODO: Open electron issue. They need to let us expose variables before any modules are loaded
+global.process = {
+	versions: {
+		chrome: ''
+	}
+};
+
 process.once('loaded', () => {
 	global.global = global;
-	global.require = require;
 	global.process = process;
-	global.Buffer = Buffer;
 });
 
 const {require: r} = remote;
 
 global.mainModules = {
 	getPort: r('get-port'),
+	/// TODO: I might be able to use the webpack externals thing here
+	electronUtil: r('electron-util'),
 	config: r('./config'),
 	portfolioUtil: r('./portfolio-util'),
 	i18n: r('./locale').i18n,
 	util: r('./util'),
+	/// TODO: Need to selectively expose `electron` sub-modules
+	electron: require('electron'),
 };
--- a/app/index.js
+++ b/app/index.js
@@ -4,6 +4,7 @@ require('strict-import')(module, {
 		'electron-debug',
 	],
 });
+const path = require('path');
 const electron = require('electron');
 const {autoUpdater} = require('electron-updater');
 const {is, disableZoom} = require('electron-util');
@@ -86,6 +87,8 @@ function createMainWindow() {
 		backgroundColor: '#1b232f', // Same as `--background-color`
 		darkTheme: isDarkMode, // GTK+3
 		webPreferences: {
+			nodeIntegration: false,
+			preload: path.join(__dirname, 'preload.js'),
 			webviewTag: false, // Disabled for security reasons since we don't use it
 			blinkFeatures: 'CSSBackdropFilter',
 		},
--- a/app/renderer/api.js
+++ b/app/renderer/api.js
@@ -1,12 +1,11 @@
 import util from 'util';
-import electron from 'electron';
 import {sha256} from 'crypto-hash';
 import PQueue from 'p-queue';
 import ow from 'ow';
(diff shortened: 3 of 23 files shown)


app/preload.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+const {remote} = require('electron');
+
+process.once('loaded', () => {
+	global.global = global;
+	global.require = require;
+	global.process = process;
+	global.Buffer = Buffer;
+});
```
sindresorhus: Took me forever before I realized I needed to expose the variables inside this event callback. I kept getting "require is not defined" and thought it was some kind of webpack issue...

kevva:
> Unless I'm misunderstanding something, this doesn't appear to have blocked much. I can still remote require stuff and most importantly still get a shell.

Hmm, yeah, that's true because we expose `require` on `global`. Why do we need that?

lukechilds:
The idea of https://github.com/hyperdexapp/hyperdex/issues/221 was to completely isolate Node.js from the render process. So in the unlikely event of an XSS that attack is sandboxed to HyperDEX and can't attack the underlying system.

I think the only way to do this is to have our own message passing either over IPC (if we can do that without exposing Node.js stuff) or over `*shudder*` HTTP. Any system functionality needed in render should be done by passing these predefined messages. That way we can be confident it can only interface with the exact system properties we've written endpoints for in the exact way we've allowed.

I know this is a lot of work but it's the only way to completely protect against system access. I'm not even saying we should definitely do that, the issues was just to consider doing it. We need more discussion on if it's even worth it.

lukechilds:
Just as an example of why I think message passing is the only solution, even if this PR correctly locked down everything apart from the whitelisted modules, it's still exposing the config module which can be manipulated into giving an attacker filesystem read/write access.

Example:

```js
const config = global.mainModules.config;
config.path = '/Users/lukechilds/.now/auth.json';
config.get().credentials[0].token;
// "My ZEIT now.sh token"
```

With message passing this wouldn't be possible because the config instance would be isolated from render in the main process. We could implement messages to get/set config params but there would obviously be no method for overwriting the path.

sindresorhus:
Ugh. I missed that. I found a way to make it work now without `require` being exposed (https://github.com/hyperdexapp/hyperdex/pull/364/commits/4524b956c9adc39d294eaed2d0eb52ecb561a45d). It was quite a rabbit hole... It pretty much makes it hard to use any external module depending on Electron.

app/preload.js
```diff
@@ -0,0 +1,29 @@
...
+global.mainModules = {
+	getPort: r('get-port'),
+	/// TODO: I might be able to use the webpack externals thing here
+	electronUtil: r('electron-util'),
+	config: r('./config'),
+	portfolioUtil: r('./portfolio-util'),
+	i18n: r('./locale').i18n,
+	util: r('./util'),
+	/// TODO: Need to selectively expose `electron` sub-modules
+	electron: require('electron'),
```
sindresorhus: We would have to handpick allowed Electron modules here as it's currently `remote.require` is still exposed, but I'll only bother if we decide to go for this solution.

sindresorhus:
Allowing only message-passing would require a *lot* of work and make the dev experience a lot more annoying...

sindresorhus:
> it's still exposing the config module which can be manipulated into giving an attacker filesystem read/write access.

Good point. We could easily fix this instance by making it a getter only, but there are probably a lot of other entry points we haven't considered.

sindresorhus:
Even message-passing is not guaranteed to be secure as it's an experimental feature: https://electronjs.org/docs/api/sandbox-option#status

lukechilds:
>but there are probably a lot of other entry points we haven't considered.

Yeah I wasn't so much pointing out that specific example, but more highlighting it as an example of how exposing any Node.js stuff directly is a potential hole that we can only really eliminate with message passing.

>Allowing only message-passing would require a lot of work and make the dev experience a lot more annoying...

I completely agree with you here, I'm not saying we should, I just opened the issue because I wanted to further discussion on it from https://github.com/hyperdexapp/hyperdex/issues/10.

That said, if we can find a middle ground where we can lock down the Node.js access a little more without completely ruining the dev experience. e.g whitelisting certain modules, I wouldn't be against that. There will likely be potential holes to wriggle through but that's obviously miles better than allowing any JS direct access to a shell.

I think it might be better to wait for a security audit until we make a decision on this.

Regarding this PR, if you can lock down `remote.require` without causing a load of friction with `require` issues I'd be happy with this. But if it's going to be a lot of work or it's going to cause dev headaches I'd say leave it for now. If it's trivial to get a system shell then locking down certain Node.js modules doesn't really give any extra security.

lukechilds:
>Even message-passing is not guaranteed to be secure as it's an experimental feature: https://electronjs.org/docs/api/sandbox-option#status

Then I'd suggest using our own local HTTP server:

>I think the only way to do this is to have our own message passing either over IPC (if we can do that without exposing Node.js stuff) or over *shudder* HTTP.

 but that's obviously a whole new can of worms. Definitely worth waiting for audit feedback before we decide how much effort is worth putting in to this.

I think there's a good chance we don't need to worry about it.

sindresorhus:
Closing this. It was a good exploration, but ultimately it's not worth it. It doesn't fully protect us but does come with a lot of annoyances. If we were to go this route, we should do it 100%.

PR #368, opened 2018-06-18
Add missing translations

kevva:
Translations for the `ForgotPassword` and `RestorePortfolio` views are still missing. There are also some in the components, as well as some that are returned from the API. Also needs to add translations for the rest of the menu items.

Probably need to settle on some sort of convention regarding the translations keys too.


app/locales/en/dashboard.json
```diff
@@ -0,0 +1,45 @@
...
+	"withdraw": {
+		"amountLabel": "Amount",
+		"confirmNetworkFee": "Confirm Network Fee",
+		"label": "Withdraw",
+		"maxAmount": "Max",
+		"networkFee": "Network Fee",
+		"recipientLabel": "Recipient",
+		"recipientPlaceholder": "Enter {{symbol}} Address",
+		"remainingBalance": "Remaining balance",
+		"successDescription": "{{amount}} {{currency}} sent to {{address}}",
```
lukechilds: `currency` should be `symbol` here. I know it's not consistent everywhere in the current code, we need to fix that. But for any new code ticker symbol should be referred to as `symbol`, and a coin/currency object should be referred to as `currency`.

kevva: Oh, that's an oversight from my side. Tried to make sure we used symbol everywhere else.

lukechilds: Yeah, super easy to get mixed up.

app/renderer/views/Exchange/Intro.js
```diff
@@ -19,15 +22,17 @@ class Intro extends React.Component {
...
+					<Trans i18n={instance} i18nKey="intro.description" t={t}>
+						<p>HyperDEX is a decentralized exchange. It has a P2P order book and trades are made via <ExternalLink url="https://en.bitcoin.it/wiki/Atomic_cross-chain_trading">cross-chain atomic swaps</ExternalLink>. This means it works slightly differently than a centralized exchange.</p>
+						<p>Orders can take a while to propagate across the P2P network. You may need to wait a while for the order book to display.</p>
+						<p>HyperDEX needs you to have correctly sized <ExternalLink url="https://bitcoin.stackexchange.com/a/61579">UTXOs</ExternalLink> to make a successful swap. For each UTXO you can make a single swap. The easiest way to ensure you can participate in a swap is to make two or three smaller deposits to your account rather than a single large deposit.</p>
+						<p>An atomic swap is comprised of six on-chain transactions. When trading a currency with high transaction fees, it&apos;s important to slightly increase your price to account for the total transaction fees. This is especially important when trading small amounts.</p>
+					</Trans>
```
sindresorhus: It's ok for now, but I don't like how we have to duplicate most of this text both here and in the translate string. It could easily get out of sync. Using numbered tags like `<0>` also makes it harder to make changes. Couldn't we, in theory, have this whole tag in the translate string and just inject it here somehow? Either at runtime or compile-time.

https://stackoverflow.com/questions/45816830/inject-jsx-formatted-string-into-react-component
https://github.com/remarkablemark/html-react-parser

Or simply just find an replace `{{i18n "intro.description"}}` in the code at compile-time.

sindresorhus:
Looking good so far 👌

app/renderer/views/Exchange/Intro.js
kevva: Yeah, I guess we can disable escaping for these cases and then turn the HTML into elements manually using `html-react-parser` or likewise.

kevva:
I've been pretty granular with the translation keys, putting them each in their respective view, which results in some duplicates. Should we put common/duplicated keys in `common.json` instead or should we just merge this and fix that later?

sindresorhus:
> Should we put common/duplicated keys in common.json instead or should we just merge this and fix that later?

Yeah, we should. Wonder if `i18next` supports referring to other keys in the translate config, so you could do:

```json
"feedback": "${common.feedback}",
```

That could also be useful if we want to reuse parts of a string.

This can be done in a follow-up PR though.

app/locale.js
```diff
@@ -8,15 +9,32 @@ i18next
...
 		initImmediate: false,
 		backend: {
 			loadPath: path.join(__dirname, 'locales', '{{lng}}', '{{ns}}.json'),
 			jsonIndent: 4,
 		},
 	});
 
+app.on('ready', () => {
+	i18next.changeLanguage(app.getLocale());
+});
```
sindresorhus: I'm seeing errors in the terminal:

```
i18next::backendConnector: loading namespace swap for language en-US failed { Error: ENOENT: no such file or directory, open '/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/locales/en-US/swap.json'
    at Object.fs.openSync (fs.js:646:18)
    at Object.module.(anonymous function) [as openSync] (ELECTRON_ASAR.js:172:20)
    at Object.fs.readFileSync (fs.js:551:33)
    at Object.fs.readFileSync (ELECTRON_ASAR.js:510:29)
    at readFile (/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/node_modules/i18next-sync-fs-backend/lib/index.js:68:27)
    at Backend.read (/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/node_modules/i18next-sync-fs-backend/lib/index.js:127:7)
    at Connector.read (/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/node_modules/i18next/dist/commonjs/BackendConnector.js:173:32)
    at Connector.loadOne (/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/node_modules/i18next/dist/commonjs/BackendConnector.js:236:10)
    at /Users/sindresorhus/dev/BarterDEX project/hyperdex/app/node_modules/i18next/dist/commonjs/BackendConnector.js:205:14
    at Array.forEach (<anonymous>)
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/locales/en-US/swap.json' }
i18next::backendConnector: loading namespace trades for language en-US failed { Error: ENOENT: no such file or directory, open '/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/locales/en-US/trades.json'
    at Object.fs.openSync (fs.js:646:18)
    at Object.module.(anonymous function) [as openSync] (ELECTRON_ASAR.js:172:20)
    at Object.fs.readFileSync (fs.js:551:33)
    at Object.fs.readFileSync (ELECTRON_ASAR.js:510:29)
    at readFile (/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/node_modules/i18next-sync-fs-backend/lib/index.js:68:27)
    at Backend.read (/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/node_modules/i18next-sync-fs-backend/lib/index.js:127:7)
    at Connector.read (/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/node_modules/i18next/dist/commonjs/BackendConnector.js:173:32)
    at Connector.loadOne (/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/node_modules/i18next/dist/commonjs/BackendConnector.js:236:10)
    at /Users/sindresorhus/dev/BarterDEX project/hyperdex/app/node_modules/i18next/dist/commonjs/BackendConnector.js:205:14
    at Array.forEach (<anonymous>)
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/sindresorhus/dev/BarterDEX project/hyperdex/app/locales/en-US/trades.json' }
```

I guess it's because `changeLanguage` expects `en` while `app.getLocale()` returns `en-US`.

sindresorhus:
@kevva I'm seeing pretty large slowdowns when switching to this branch. Especially noticeable when typing in input fields or switching between the main views. Can you look into that?

kevva:
I guess it's because it'll run the `t` function on every render. That's one of the advantages of using the HOC. It'll prevent the translations from being ran again by [doing a comparison](https://github.com/i18next/react-i18next/blob/master/src/translate.js#L34-L40) in `shouldComponentUpdate`.

sindresorhus:
I'm surprised calling maybe 20 `t()` methods can add almost 1 second render time. Transforming the string should not be that slow. It kinda feels like it's still doing IPC communication on render somehow... We should memoize `t()` regardless.

sindresorhus:
@kevva Can you fix the merge conflict?

sindresorhus:
@kevva If you open the Devtron tab in DevTools, then `IPC` in the sidebar, press Record, and then go to a new main tab in the UI, you'll a gadzillion IPC calls.

kevva:
> @kevva If you open the Devtron tab in DevTools, then IPC in the sidebar, press Record, and then go to a new main tab in the UI, you'll a gadzillion IPC calls. Don't wait too long until you press Record again to stop, otherwise DevTools will hang.

Yup, it's due to https://github.com/electron/electron/blob/master/docs/api/remote.md#remote-objects. I decided to use the `ipc` module instead which seems to fix the issue.

app/locale.js
kevva: Yes. It falls back to `en` if `en-US` isn't found. It's because of the `debug` flag.

sindresorhus:
Merging as this looks good now and the performance issue has been fixed 👌

app/locale.js
sindresorhus: I think we should use full language codes though to be explicit, so `en-US` instead of `en`. Same with the other ones. Crowdin also uses full language codes.

sindresorhus:
@kevva Can you fix https://github.com/hyperdexapp/hyperdex/pull/368#discussion_r198012393 ?

I also noticed in the DevTools console:

> i18next::translator: missingKey en-US dashboard activity.swappedTitle activity.swappedTitle
> logger.js:23 i18next::translator: missingKey en-US dashboard activity.swappedDescription activity.swappedDescription

sindresorhus:
> Should we put common/duplicated keys in common.json instead or should we just merge this and fix that later?

@kevva Can you open an issue about this so we don't forget?

app/locale.js
kevva: Then we have to maintain languages like `en-GB` too. Only language it makes sense for is Chinese imo.

sindresorhus:
> @kevva Can you fix #368 (comment) ?

Forgot to say, I already fixed this here as I needed it for Crowdin: https://github.com/hyperdexapp/hyperdex/commit/ef47f50085a6c42638c8bf2add0edcfd398a2be6

sindresorhus:
@kevva I noticed some more places with missing translations:

- https://github.com/hyperdexapp/hyperdex/blob/a7e56068a5d8098d31e0ed6005456b58f0a6250e/app/renderer/containers/CreatePortfolio.js#L46
- https://github.com/hyperdexapp/hyperdex/blob/a7e56068a5d8098d31e0ed6005456b58f0a6250e/app/renderer/containers/RestorePortfolio.js#L49
- https://github.com/hyperdexapp/hyperdex/blob/a7e56068a5d8098d31e0ed6005456b58f0a6250e/app/renderer/containers/ForgotPassword.js#L43
- https://github.com/hyperdexapp/hyperdex/blob/a7e56068a5d8098d31e0ed6005456b58f0a6250e/app/renderer/containers/ForgotPassword.js#L17
- https://github.com/hyperdexapp/hyperdex/blob/a7e56068a5d8098d31e0ed6005456b58f0a6250e/app/renderer/containers/Exchange.js#L129-L130
- https://github.com/hyperdexapp/hyperdex/blob/a7e56068a5d8098d31e0ed6005456b58f0a6250e/app/renderer/components/DepthChart.js#L151
- Many strings in https://github.com/hyperdexapp/hyperdex/blob/a7e56068a5d8098d31e0ed6005456b58f0a6250e/app/renderer/components/SwapDetails.js
- https://github.com/hyperdexapp/hyperdex/blob/a7e56068a5d8098d31e0ed6005456b58f0a6250e/app/renderer/views/Exchange/Order.js#L327
- We need to translate `statusFormatted` in https://github.com/hyperdexapp/hyperdex/blob/d2d463ba50470f40a4664abee00319abe0e7c35f/app/renderer/swap-db.js

sindresorhus:
I'm also getting an error in the console:

```
warning.js:33 Warning: Failed context type: The context `i18n` is marked as required in `Trans`, but its value is `undefined`.
    in Trans (created by Intro)
    in main (created by Modal)
    in div (created by Modal)
    in div (created by Modal)
    in Modal (created by Intro)
    in div (created by Intro)
    in Intro (created by Consumer)
    in Consumer (created by Subscribe)
    in Subscribe (created by Exchange)
    in Exchange (created by View)
    in View (created by AppView)
    in AppView (created by Consumer)
    in Consumer (created by Subscribe)
    in Subscribe (created by App)
    in App (created by HotExportedApp)
    in AppContainer (created by HotExportedApp)
    in HotExportedApp
    in Provider (created by Consumer)
    in Consumer (created by Provider)
    in Provider
```

sindresorhus:
I also noticed this:

screen shot 2018-06-27 at 18 37 12

sindresorhus:
screen shot 2018-06-27 at 22 03 08

kevva:
> I'm also getting an error in the console

That is a bug in `react-i18next`. It shouldn't depend on `i18n` being in context since you can pass it in as a prop.

sindresorhus:
> That is a bug in react-i18next. It shouldn't depend on i18n being in context since you can pass it in as a prop.

Can you open an issue on `react-i18next` about that?

PR #370, opened 2018-06-19
CSP improvements

sindresorhus:
- We now apply the rules to the whole app, so it applies even to new windows.
- Since we apply it to the whole app, there's no way to trick our app into redirecting and then go around our CSP.
- I also added some additional rules that are not covered by the `default-src 'none';` rule.
- ~~Since it's dynamically injected now, we can make it even stricter and only allow the local port marketmaker is running on.~~
- The rules are now also more readable.


lukechilds:
Great idea generating it dynamically.

I also just realised, our CSP will block remote `marketmaker` nodes. Not worth blocking this for that as basically nobody apart from me uses that. And even if someone did use a remote node, they would likely tunnel it locally rather than actually access it over the internet so would probably still work. But just worth baring in mind.

We could dynamically add the custom `marketmaker` url if it's set to fix this, but really not worth the effort right now.

lukechilds:
Oh, just realised you're now injecting headers rather than using the meta tag so maybe not so simple to change this.

sindresorhus:
Done

Issue #262, opened 2018-05-21
[Improvement]


Erlix322:
Could you add a cancel Button in the Exchange view?
Would be great to cancel a pending order directly in there.

---

HyperDEX 0.1.0-alpha.1
Electron 2.0.0
win32 10.0.16299
Locale: de

sindresorhus:
Thanks for the suggestion. We plan to add this.

lukechilds:
Currently this functionality isn't very useful due to the fact that you can only cancel pending swaps. Normally a swap will either be matched fairly quickly or never match.

Cancelling swaps that never match obviously doesn't add any real usability. Cancelling a swap before it's matched is difficult as it gets matched so fast. And cancelling a swap after it's matched would be useful for the user and is probably what most users expect to be able to do but isn't allowed by `marketmaker`.

I raised this before with James and asked if we could cancel matched swaps but he was strongly opposed to it. Is there any chance you would reconsider @jl777?

jl777:
we could slow the ordermatch down a lot, but that seems not so good.

alternatively, changing the protocol would be required. What do you suggest to be done? Alice would need to force bob to reclaim the payment and deposit. I am not against it, it just seems quite impractical

lukechilds:
>we could slow the ordermatch down a lot, but that seems not so good.

No, fast matching is good.

>alternatively, changing the protocol would be required. What do you suggest to be done? Alice would need to force bob to reclaim the payment and deposit. I am not against it, it just seems quite impractical

@jl777 couldn't you just stop Alice from responding to Bob anymore? So as far as Bob's concerned the swap failed due to network/communication issues. Then the swap will unwind/reclaim using the normal procedure.

Obviously id there's any way to instantly cancel and reclaim funds that's great, but I'm guessing that's not possible due to the timelock.

If so we will display a clear message to the user when they click cancel. We'll let them know that they'll lose their dexfee and it may take up to 4 hours for their funds to be reclaimed. We'll require them to confirm they're ok with that before actually cancelling.

jl777:
user can effectively do this just by exiting, but I see how it is good to do it your way. I would need to add external signalling to the main swap threads. If we can limit to one cancel at a time, that would enable a global variable based method, it would then generate a "swapcancelled" message so you would know when you can cancel another one

jl777:
but you want to cancel swaps that are already in progress, dont you?
there is no limit to the number of those, other than the practical limit of how many you can get going before they start completing

[...]

PR #371, opened 2018-06-19
Make the swap list "Cancel" button red

sindresorhus:
To make it less likely for people to click it accidentally when they want the "View" button.

screen shot 2018-06-19 at 18 51 46

### DIFF
--- a/app/renderer/components/SwapList.scss
+++ b/app/renderer/components/SwapList.scss
@@ -169,6 +169,10 @@
 			}
 		}
 	}
+
+	.cancel :not([disabled]) {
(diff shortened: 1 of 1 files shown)


lukechilds:
Do you think we should remove the cancel button altogether?

It doesn't seem like the functionality is going to be fixed until `marketmaker` v2: https://github.com/hyperdexapp/hyperdex/issues/262#issuecomment-396587751

sindresorhus:
> Do you think we should remove the cancel button altogether?

Yes, I don't think we should provide "broken" functionality.

---

I still like to have this merged so it's red when we re-enable the Cancel button when using mm v2.

PR #372, opened 2018-06-19
ERC20 support

lukechilds:
This _should_ work, but I seem to be getting errors doing ERC20 withdrawals. I think this is purely an issue with the ETH node but I'll test this more tomorrow.

The main difference is that now if an ERC20 token is used the withdraw value is USD is calculated from the ETH USD price not the token USD price (because ERC20 tx fees are paid in ETH).

screen shot 2018-06-19 at 8 00 35 pm

I also added Unicoin as a test token. We must specify the decimals in the coins file when adding ERC20 tokens otherwise there's a bug in `marketmaker` that will cause it to die under certain circumstances.

Also now log the TXID on withdraw for debugging.


lukechilds:
Should probably wait for most of these to be resolved before merging:

https://github.com/artemii235/SuperNET/issues/84
https://github.com/artemii235/SuperNET/issues/71
https://github.com/artemii235/SuperNET/issues/83
https://github.com/artemii235/SuperNET/issues/82
https://github.com/artemii235/SuperNET/issues/80
https://github.com/artemii235/SuperNET/issues/81

app/renderer/views/Dashboard/WithdrawModal.js
```diff
@@ -51,18 +54,24 @@ class WithdrawModal extends React.Component {
...
+		const txFeeUsd = formatCurrency(txFee * cmcPriceUsd);
+
+		this.setState({txFeeCurrencySymbol, txFee, txFeeUsd, broadcast});
 	};
 
 	confirmButtonHandler = async () => {
 		this.setState({isBroadcasting: true});
-		const {amount, currency, address} = await this.state.broadcast();
+		const {txid, amount, symbol, address} = await this.state.broadcast();
+		console.log({txid, amount, symbol, address});
```
kevva: Accidental `console.log`?

app/renderer/api.js
```diff
@@ -328,8 +328,12 @@ export default class Api {
 				broadcast: 1,
 			});
 
+			if (response.error) {
+				throw errorWithObject('Couldn\'t create withdrawal transaction', response);
```
kevva: Out of scope for this PR, but maybe this should be converted to an `ApiError` or something?

app/renderer/views/Dashboard/WithdrawModal.js
lukechilds: I believe me and @sindresorhus discussed this and decided we should log TXIDs to the console (we don't do anything with them yet in the notification).

So just decided to log all the TX data fro context.

sindresorhus: Until we get it in the UI, we should keep it for debugging purposes.

PR #381, opened 2018-06-25
Stop using deprecated React lifecycle hooks

### LINKED ISSUE #373: Use React strict mode
https://reactjs.org/docs/strict-mode.html

This will also let us remove this: https://github.com/hyperdexapp/hyperdex/blob/60d9feecda1449222ac914f92e247b6e2cf54957/package.json#L102

Assigning to @kevva since he has the most React experience.

kevva:
We have some third party components that causes some noise when using strict mode, but our components should be okay at least.

Fixes #373.

### DIFF
--- a/app/renderer/components/Modal.js
+++ b/app/renderer/components/Modal.js
@@ -11,19 +11,29 @@ class Modal extends React.Component {
 		delay: 400,
 	};
 
+	static getDerivedStateFromProps(props, state) {
+		if (!state.open && props.open) {
+			return {
+				isOpen: true,
+				animationType: 'open',
+			};
+		}
+
+		if (state.open && !props.open) {
+			if (props.onClose) {
+				props.onClose();
+			}
+
+			return {animationType: 'close'};
+		}
+	}
+
 	state = {
 		isOpen: false,
 		animationType: 'close',
 	};
 
 	elementRef = React.createRef();
-
-	openHandler = () => {
-		this.setState({
-			isOpen: true,
-			animationType: 'open',
-		});
-	};
 
 	closeHandler = () => {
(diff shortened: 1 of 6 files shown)


sindresorhus:
Let's not enable the Strict mode itself for now. It's very noisy. Just add a `TODO` comment about it to enable it before upgrading to React 17. The main offenders are `react-select` and `recharts`, which we can't easily improve, so better to just let the community them when React actually removes those methods.

Also @kevva is fixing `styled-jsx`: https://github.com/zeit/styled-jsx/pull/457

sindresorhus:
I got this error when trying to withdraw some DOGE in the Withdraw modal:

```
Warning: Modal.getDerivedStateFromProps(): A valid state object (or null) must be returned. You have returned undefined.
```

kevva:
> I got this error when trying to withdraw some DOGE in the Withdraw modal:

Didn't take into account when `state` hasn't changed in the `Modal` component.

sindresorhus:
@kevva The modal doesn't close anymore after a successful withdrawal.

app/renderer/components/Modal.js
```diff
@@ -11,20 +11,32 @@ class Modal extends React.Component {
 		delay: 400,
 	};
 
+	static getDerivedStateFromProps(props, state) {
+		if (!state.open && props.open) {
```
kevva: Lol, `state.open` should be `state.isOpen`.

kevva:
@sindresorhus, fixed now. Besides a typo (`state.open` instead of `state.isOpen`), we also need to handle when `isOpen` is still `true` but the modal is closing.

sindresorhus:
Works perfectly now :)

PR #387, opened 2018-06-28
Add further translations

kevva:
This should address the missing translations mentioned in https://github.com/hyperdexapp/hyperdex/pull/368#issuecomment-400629256.

### DIFF
--- a/app/locales/en-US/swap.json
+++ b/app/locales/en-US/swap.json
@@ -1,19 +1,41 @@
 {
 	"tempMessageDescription": "Simple atomic swaps with minimal user interaction",
 	"tempMessageTitle": "Coming soon…",
+	"timedOut": "Swap timed out",
 	"title": "Swap",
 	"details": {
+		"alicepayment": "Alice Payment",
+		"alicespend": "Alice Spend",
+		"bobdeposit": "Bob Deposit",
+		"bobpayment": "Bob Payment",
+		"broadcast": "Broadcast",
+		"buy": "Buy",
 		"copyDebugData": "Copy Swap Debug Data",
+		"exchanging": "Exchanging",
+		"executed": "Executed",
 		"for": "For",
+		"myfee": "My Fee",
 		"order": "Order",
+		"requested": "Requested",
+		"pending": "Pending",
 		"price": "Price",
+		"sell": "Sell",
 		"stats": "The executed price was {{percentCheaperThanRequested}}% cheaper than requested!",
 		"transactions": "Transactions",
+		"youExchanged": "You exchanged",
+		"youReceived": "You received",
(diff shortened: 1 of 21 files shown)


sindresorhus:
🙌

app/locales/en-US/swap.json
```diff
@@ -1,19 +1,41 @@
 {
 	"tempMessageDescription": "Simple atomic swaps with minimal user interaction",
 	"tempMessageTitle": "Coming soon…",
+	"timedOut": "Swap timed out",
 	"title": "Swap",
 	"details": {
+		"alicepayment": "Alice Payment",
```
sindresorhus: @lukechilds @kevva Shouldn't this be without the name (`Alice`)?

lukechilds: Alice and Bob are important distinctions. Alice is the user and Bob is the market maker node the user is trading with.

Maybe we could change `Alice => My` and `Bob => Peer`?.

Or to just totally abstract the whole thing:

```js
{
	"myfee": "My Fee",
	"bobdeposit": "Seller Deposit",
	"alicepayment": "My Payment",
	"bobpayment": "Seller Payment",
	"alicespend": "Claim Seller Payment",
}
```

Although I'm not too keen on that as "seller" is confusing because it always refers to Bob, even if you've placed a sell order. Although I suppose we could toggle between using the word `Seller | Buyer` depending on whether it's a buy or sell order.

sindresorhus:
Sorry for being unclear. I just meant that we should not make the name localizable. We should still use `Alice Payment`, but only localize the `Payment` part. `Alice` can be hard-coded in the component.

lukechilds:
Oh, I get you, yeah, it doesn't make sense to translate the Alice/Bob names into different languages.

Although now might be a good time to abstract these into more user friendly titles. What do you think of:

```js
{
	"myfee": "My Fee",
	"bobdeposit": "Seller Deposit",
	"alicepayment": "My Payment",
	"bobpayment": "Seller Payment",
	"alicespend": "Claim Seller Payment",
}
```

sindresorhus:
`My` => `Your`? Not a big fan of me-ifying interfaces.

And `Claim Seller` => `Seller Claim`?

lukechilds:
>`My` => `Your`?

Yeah, that sounds better.


> `Claim Seller` => `Seller Claim`

I don't think that makes sense, it's not "The seller claiming a payment" but "The user claiming the sellers payment".

app/locales/en-US/swap.json
kevva: Yes, thought about that at first, but since we use the Alice and Bob distinction it made sense to have them in the translation. But yeah, we can make it a variable instead.

PR #389, opened 2018-06-29
Add dashboard intro

lukechilds:
Hopefully this will notify all users of the risks and make them aware of the maturity of the software.

MIT license should cover us for any legal disputes but @pondsea wanted me to explicitly bring it up here too. I give a breif summary about no warranty/liability and state that "By using HyperDEX you are agreeing to the terms of the MIT License".

I think that should be more than enough.

screen shot 2018-06-29 at 3 10 01 pm

The screenshot was taken with localisation disabled. Using the localisation files the text gets a bit garbled. I tried to copy the `<n>` pattern I'd seen in other localisation files but it doesn't look like I got it right. Is this a standardised format? Is there an automatic conversion tool we can use instead of manually creating the `<n>` tags.

Also, suggestions for text modifications welcome.

### DIFF
--- a/app/locales/en-US/dashboard.json
+++ b/app/locales/en-US/dashboard.json
@@ -1,7 +1,7 @@
 {
 	"intro": {
 		"title": "✨ Thanks for Helping Us Test HyperDEX!",
-		"description": "<0>Please keep in mind <1>this project is still in Alpha</1>.</0><1>Although loss of funds is unlikely, you should only trade with real currency if you can take that risk.</1><2>If you just want to test HyperDEX, try trading between the test currencies <1>BEER</1> and <2>PIZZA</2> instead. You can get free <3>BEER</3> at the <4><1>BEER</1> faucet</4>.</2><3>HyperDEX is Open Source software provided \"as is\", without warranty of any kind. The authors or copyright holders are not liable for any damages caused as a result of using this software.</3><4>By using HyperDEX you are agreeing to the terms of the <1>MIT License</1>.</4>"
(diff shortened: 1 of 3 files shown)


sindresorhus:
> Is there an automatic conversion tool we can use instead of manually creating the <n> tags.

Not sure, but we definitely need that. See: https://github.com/hyperdexapp/hyperdex/issues/390

sindresorhus:
> I tried to copy the <n> pattern I'd seen in other localisation files but it doesn't look like I got it right.

I think the numbers have to be unique for each tag.

sindresorhus:
Travis is failing because of some lint error.

sindresorhus:
The text looks good.

lukechilds:
Hmmn, I tried using unique tag names and it messed it up even more 🤔.

Lint issue fixed.

Do you know what the actual format is called, like is it a i18n standard thing? I can't find any information on it.

sindresorhus:
I think it's just specific to the `react-i18next` module. Not much info in the docs either: https://react.i18next.com/components/trans-component

sindresorhus:
> Hmmn, I tried using unique tag names and it messed it up even more 🤔.

What does it look like messed up? I'm not able to reproduce that.

lukechilds:
I see this:

screen shot 2018-06-29 at 10 21 50 pm

kevva:
Best way to get it right is to just wrap it in the `Trans` component without adding a translation for it. Then it'll report in the console that a translation is missing, and print it with the correct formatting.

lukechilds:
Cheers @kevva That sorted it.

lukechilds:
I think this should be good to merge now. Then we can get on translations ASAP.

PR #400, opened 2018-07-04
Light theme ✨

sindresorhus:
I've tried to reuse CSS variables as much as possible to create the least amount of light theme specific styles as possible. I've also made it possible to create additional themes in the future if wanted. For example, creating a theme that just uses different colors would not be difficult at all.

This work would have been much simpler if [CSS had color modifiers](https://twitter.com/sindresorhus/status/991610996589346817)...

screen shot 2018-07-04 at 15 27 12

screen shot 2018-07-04 at 15 31 16


lukechilds:
Also, too small issues I noticed.

The % of your portfolio holdings isn't visible in the dashboard.

Original:

screen shot 2018-07-04 at 4 27 53 pm

Emphasised:

screen shot 2018-07-04 at 4 28 15 pm

And the Debug button in the swap modal still has the dark button styles:

screen shot 2018-07-04 at 4 28 59 pm

sindresorhus:
Hmm, feels a bit too bright.

lukechilds:
Ok.

Good work on refactoring into CSS variables btw. Getting the light theme looking perfect isn't crucial right now. As long as it's implemented for launch and easily tweakable with the vars that's the main thing.

We can get some new visuals done based on this and finalise it at a later date when we're less pressed for time. 👌 

One those points in my last comment are resolved I'm happy for this to be merged.

lukechilds:
Oh also, two other minor things.

How do you feel about a custom dropdown for the theme selection? The default dropdown styles look good on the light theme but look a bit out of place on the dark theme:

screen shot 2018-07-04 at 5 15 04 pm

And I think the style changes may have broke the login button in dark mode, I don't remember the "Login" text being dark before:

screen shot 2018-07-04 at 5 15 15 pm

sindresorhus:
> The % of your portfolio holdings isn't visible in the dashboard.

Fixed

> And the Debug button in the swap modal still has the dark button styles:

Fixed

> How do you feel about a custom dropdown for the theme selection? The default dropdown styles look good on the light theme but look a bit out of place on the dark theme:

I agree about the control itself, but I think it should be the native menu when opened. Let's preserve the little native feeling we have left when possible.

screen shot 2018-07-05 at 18 04 34
screen shot 2018-07-05 at 18 04 39



> And I think the style changes may have broke the login button in dark mode, I don't remember the "Login" text being dark before:

That's an intentional change. Dark mode has so much white text on black background, I think this makes it look more interesting and make the main button stand out more.

lukechilds:
Looks great.

>I agree about the control itself, but I think it should be the native menu when opened. Let's preserve the little native feeling we have left when possible.

👌

sindresorhus:
Travis failing is unrelated to this PR. Seems to be some kind of race issue with Babel plugins.

PR #402, opened 2018-07-04
Improve `fire-every` module

### LINKED ISSUE #401: Use time argument for `fireEvery()` as first argument and allow passing in an object or a timestamp
Currently we have:
```js
const ONE_HOUR = 1000 * 60 * 60;
fireEvery(async () => {
  // Do some stuff
  await this.kickstartStuckSwaps();
  // Do some more stuff
}, ONE_HOUR);
```

This is much more readable:
```js
const ONE_HOUR = 1000 * 60 * 60;
fireEvery(ONE_HOUR, async () => {
  // Do some stuff
  await this.kickstartStuckSwaps();
  // Do some more stuff
});
```

We should also allow passing in an object for better readability:
```js
fireEvery({hour: 1}, async () => {
  // Do some stuff
  await this.kickstartStuckSwaps();
  // Do some more stuff
});
```

We could add an inverse option to [parse-ms](https://github.com/sindresorhus/parse-ms) to handle converting this to a timestamp.

sindresorhus:
- Nicer way to specify the interval
- Make the interval the first argument
- Add option to start after delay

Fixes #401

### DIFF
--- a/app/renderer/fire-every.js
+++ b/app/renderer/fire-every.js
@@ -1,12 +1,14 @@
 import toMilliseconds from '@sindresorhus/to-milliseconds';
 import delayModule from 'delay';
 
-const fireEvery = async (delay, cb, options = {}) => {
+const fireEvery = async (delay, cb, options) => {
 	if (typeof delay !== 'number') {
 		delay = toMilliseconds(delay);
 	}
 
-	if (options.startAfterDelay) {
+	options = Object.assign({fireInstantly: true}, options);
+
+	if (!options.fireInstantly) {
 		await delayModule(delay);
 	}
 
--- a/app/renderer/fire-every.js
+++ b/app/renderer/fire-every.js
@@ -1,4 +1,15 @@
-const fireEvery = async (cb, delay) => {
+import toMilliseconds from '@sindresorhus/to-milliseconds';
+import delayModule from 'delay';
+
+const fireEvery = async (delay, cb, options = {}) => {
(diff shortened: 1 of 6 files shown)


app/renderer/fire-every.js
```diff
@@ -1,4 +1,15 @@
-const fireEvery = async (cb, delay) => {
+import toMilliseconds from '@sindresorhus/to-milliseconds';
+import delayModule from 'delay';
+
+const fireEvery = async (delay, cb, options = {}) => {
+	if (typeof delay !== 'number') {
+		delay = toMilliseconds(delay);
+	}
+
+	if (options.startAfterDelay) {
```
sindresorhus: I'm not happy about this option name. Any suggestions?

Lodash documents the debounce method as follows:

> [options.trailing=true] (boolean): Specify invoking on the trailing edge of the timeout.

So we could maybe call it `trailing`.

lukechilds: I would suggest `options.instant = true` as the default.

Setting `options.instant = false` would wait until after the delay for the initial fire.

lukechilds: or `fireInstantly` if you want something more descriptive.

kevva: > So we could maybe call it `trailing`

I've always found those option names in Lodash very confusing.

> I would suggest options.instant = true as the default.

This is more clear to me. Or `startInstant` or something.

sindresorhus:
Done

PR #407, opened 2018-07-05
Add ability to rename and delete portfolio

### LINKED ISSUE #271: Where i can find the private key?
Hello,
where i can find the private key of Bitcoin address or any other coin ?
thanks

kevva:
Rename portfolio input:

screen shot 2018-07-10 at 02 25 42

Deletion confirmation modal:

screen shot 2018-07-10 at 02 17 36

View seed phrase password input modal:

screen shot 2018-07-10 at 02 18 20

View seed phrase modal:

screen shot 2018-07-10 at 02 18 26

---

Fixes #271


sindresorhus:
> Will investigate a way to add them automatically without having to manually write them in every JSON file.

I think Crowdin will do this for us. You just add the English and Crowdin will do a PR with new translations when available.

app/renderer/views/Settings/SeedPhraseModal.js
```diff
@@ -0,0 +1,109 @@
...
+		this.setState({isOpen: true});
+	}
+
+	handleClose = () => {
+		this.setState({
+			isOpen: false,
+			passwordError: null,
+			passwordInputValue: '',
+			seedPhrase: '',
+		});
```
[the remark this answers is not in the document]
kevva: Any reason not to use the `SuperContainer` component instead?

app/renderer/containers/App.js
```diff
@@ -117,10 +118,25 @@ class AppContainer extends Container {
 		});
 	}
 
+	async verifyPassword(password) {
```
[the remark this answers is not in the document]
kevva: Yes, named it after its context at first. Good catch.

app/renderer/views/Settings/SeedPhraseModal.js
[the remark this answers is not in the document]
kevva: I went with a getter instead. It's easier to read and understand and only requires minimally more code:

```js
class Component extends React.Component {
	state = this.initialState;

	get initialState() {
		return {
			isOpen: false
		}
	}

	componentWillUnmount() {
		this.setState(this.initialState);
	}
}

class Component extends SuperComponent {
	getInitialState = () => ({
		isOpen: false
	});

	componentWillUnmount() {
		this.resetState();
	}
}
```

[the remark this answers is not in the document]
kevva: I went with a getter instead. It's easier to read and understand and only requires minimally more code:

```js
class Component extends React.Component {
	state = this.initialState;

	get initialState() {
		return {
			isOpen: false
		}
	}

	componentWillUnmount() {
		this.setState(this.initialState);
	}
}

class Component extends SuperComponent {
	getInitialState = () => ({
		isOpen: false
	});

	componentWillUnmount() {
		this.resetState();
	}
}
```

app/renderer/views/Settings.js
```diff
@@ -42,7 +45,12 @@ class Settings extends React.Component {
 						<main>
 							<div className="section">
 								<h3>{t('portfolio')}</h3>
+								<RenamePortfolio/>
 								<CurrencySetting/>
+								<div className="form-group">
```
sindresorhus: This group needs a subheader. Right now it visually looks like it's part of "Enabled currencies".

app/renderer/views/Settings.scss
```diff
@@ -41,6 +41,15 @@
 			color: var(--text-color);
 			line-height: 1.5;
 		}
+
+		.modal-wrapper {
+			display: inline-block;
```
sindresorhus: Let's prefer flex as much as possible. `inline-block` => `inline-flex`.

app/locales/en-US/settings.json
```diff
@@ -1,8 +1,19 @@
 {
 	"app": "App",
+	"deletePortfolio": {
+		"confirm": "I understand, delete this portfolio",
+		"description": "<0>Unless you save your seed phrase, this action cannot be undone and will permanently delete the <1><0>{{name}}</0></1> portfolio.</0><1>Please type in the name of the portfolio to confirm.</1>",
+		"title": "Delete portfolio"
```
sindresorhus: `Delete portfolio` => `Delete Portfolio`

And minor, but I would order the keys in the order they appear on the screen, from top to bottom.

sindresorhus:
If I click "View Seed Phrase", enter my password, Submit, I can no longer press Esc to close the modal. I can close the modal with Esc when in the "Enter password" view.

sindresorhus:
When viewing my seed phrase and I press the `X` button, I can see it quickly switching back to the "Enter password" view before animating that out.

sindresorhus:
I don't think the portfolio ID should be changed on portfolio rename, only the portfolio name. @lukechilds Do you agree?

app/renderer/views/Settings/RenamePortfolio.js
```diff
@@ -0,0 +1,38 @@
...
+const {renamePortfolio} = remote.require('./portfolio-util');
+const t = translate('settings');
+
+class RenamePortfolio extends React.Component {
+	state = {
+		isUpdating: false,
+	};
+
+	handleChange = async newName => {
+		this.setState({isUpdating: true});
```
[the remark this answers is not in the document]
sindresorhus: I think the actual action of changing the portfolio name should be debounced to 400 or 500ms. This is quite common for persisting inputs in React.

app/renderer/views/Settings/RenamePortfolio.js
```diff
@@ -0,0 +1,38 @@
...
+		});
+
+		this.setState({isUpdating: false});
+	};
+
+	render() {
+		return (
+			<div className="form-group">
+				<label>{t('name')}</label>
+				<Input value={appContainer.state.portfolio.name} readOnly={this.state.isUpdating} onChange={this.handleChange}/>
```
sindresorhus: It should never be `readOnly`. It should just use debouncing instead.

app/renderer/views/Settings.js
kevva: What should it say?

app/renderer/views/Settings/SeedPhraseModal.js
```diff
@@ -0,0 +1,107 @@
...
+		} catch (err) {
+			console.error(err);
+
+			this.setState({
+				isVerifying: false,
+				passwordInputValue: '',
+				passwordError: err.message,
+			});
+
+			this.passwordInputRef.current.focus2();
```
sindresorhus: `focus2`?

kevva:
> If I click "View Seed Phrase", enter my password, Submit, I can no longer press Esc to close the modal. I can close the modal with Esc when in the "Enter password" view.

The modal loses it's focus on submitting the form which causes the `onKeyUp` event handler to not listen. We would need access the `ref` for the modal and update it outside, or somehow update it automatically in `Modal`, but that would probably involve comparing children to new children.

> I don't think the portfolio ID should be changed on portfolio rename, only the portfolio name. @lukechilds Do you agree?

I agree. Should be fixed.

> Renaming currently allows setting the name to empty, this should not be allowed. If the user removes everything, we should just revert to the original name on blur.

Yup, fortunately there was `didClose`. I reset the state there instead.

> Renaming currently allows setting the name to empty, this should not be allowed. If the user removes everything, we should just revert to the original name on blur.

Do we even need to do any renaming actions until blur? That way we can skip running `renamePortfolio` on change (even with the debounce).

sindresorhus:
> But there are modules like davidtheclark/focus-trap-react that handles this.

Yeah, let's use that one.

> > Renaming currently allows setting the name to empty, this should not be allowed. If the user removes everything, we should just revert to the original name on blur.

> Yup, fortunately there was didClose. I reset the state there instead.

You used the wrong quoted paragraph above, but I get what you're referring to.

> Do we even need to do any renaming actions until blur? That way we can skip running renamePortfolio on change (even with the debounce).

I think persisting it on blur is fine in this case.

app/renderer/views/Settings/SeedPhraseModal.js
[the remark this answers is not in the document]
sindresorhus: Your way will not work if a component sets state that is not defined in the `initialState`, which might lead to hard to debug issues. Not a problem in this specific case, but if you use your pattern in other places, it might become an issue. I also don't see the problem with abstracting a common need. IMHO, resetting state is something that should have been included in `React.Component` by default.

I also have plans to be able to hook component state into `unstated-debug` so we can debug this state too, so we'll have to subclass at some point anyway.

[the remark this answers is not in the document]
sindresorhus: It also has slightly different behavior. Your version will call `initialState()` at the import time, while my version will call `getInitialState()` at construction time `new Component()`. Imagine a scenario where the state is not yet ready at import time.

sindresorhus:
I commented on https://github.com/atomiclabs/hyperdex/pull/407#discussion_r202310184, in case you don't see it.

app/renderer/views/Settings.js
```diff
@@ -9,6 +9,9 @@ import {instance, translate} from '../translate';
 import TabView from './TabView';
```
lukechilds: This file `app/renderer/views/Settings.js` should probably be moved to `app/renderer/views/Settings/index.js` to match the same pattern as the other complex views.

app/renderer/views/Settings.js
lukechilds: I agree they look out of place but it's also quite easy to pick a title for them.

Maybe it would be better to move the buttons underneath the rename portfolio input. They are semi related.

screen shot 2018-07-13 at 6 14 03 pm

lukechilds:
This is looking great @kevva.

There is a slight style issue with the password error:

screen shot 2018-07-13 at 6 01 22 pm

app/renderer/views/Settings.js
lukechilds: Or if we go with a title "Actions" is the only thing I can think of that makes sense but it's not a great title.

sindresorhus:
@lukechilds That is really strange. Can you try restarting HyperDEX and see if the issue is still there? Might be an HMR bug.

This is what I get:

screen shot 2018-07-13 at 18 41 45

kevva:
@lukechilds , that definitely looks like the CSS isn't loaded correctly. The font size and margin should be handled. Try what @sindresorhus said.

app/renderer/views/Settings.js
kevva: Or `Manage`?

app/renderer/views/Settings/SeedPhraseModal.js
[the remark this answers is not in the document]
kevva: > Your way will not work if a component sets state that is not defined in the initialState, which might lead to hard to debug issues.

That's a good point.

app/renderer/views/Settings.js
sindresorhus: > Or Manage?

👍

sindresorhus:
Looks good to me when https://github.com/atomiclabs/hyperdex/pull/407#discussion_r202375420 is done.

sindresorhus:
Nice work, @kevva :)

app/renderer/views/Settings.js
sindresorhus: Done in https://github.com/atomiclabs/hyperdex/commit/d3f3a1adaa066460a503954fa8d6b2a5cf1cc917

PR #409, opened 2018-07-05
Add ChainZilla (ZILLA) currency

sindresorhus:
Taken from: https://github.com/jl777/coins/commit/c9fb1711a47971a07ce53cc010f5f0bf19a1f3c3

### DIFF
--- a/app/marketmaker/supported-currencies.js
+++ b/app/marketmaker/supported-currencies.js
@@ -937,6 +937,26 @@ const supportedCurrencies = [
 			},
 		],
 	},
+	{
+		coin: 'ZILLA',
+		name: 'ChainZilla',
+		asset: 'ZILLA',
+		rpcport: 10041,
+		electrumServers: [
+			{
+				host: 'electrum1.cipig.net',
+				port: 10028,
+			},
+			{
+				host: 'electrum2.cipig.net',
+				port: 10028,
+			},
+			{
+				host: '158.69.0.198',
+				port: 10028,
+			},
+		],
+	},
 ];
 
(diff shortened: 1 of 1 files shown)


app/marketmaker/supported-currencies.js
```diff
@@ -937,6 +937,26 @@ const supportedCurrencies = [
...
+			{
+				host: 'electrum1.cipig.net',
+				port: 10028,
+			},
+			{
+				host: 'electrum2.cipig.net',
+				port: 10028,
+			},
+			{
+				host: '158.69.0.198',
```
lukechilds: Do we know who runs `158.69.0.198`? Just because it's been added to the `jl777/coins` repo doesn't mean we should trust it as safe.

A malicious Electrum server operator could broadcast false information about a swap they're involved in, telling us they've broadcast a payment TX while simultaneously spending those UTXOs on the real chain, making the TX invalid. We trust the Electrum response and continue with the atomic swap protocol, releasing our funds to the other peer.

Whilst we're connected to the malicious Electrum server, everything looks ok, however if you connect to an honest server you'll see the real state of the blockchain which is that we have sent funds but haven't received any.

We need to be very careful which Electrum servers we trust. `cipig.net` is run by Cipi, a Komodo Core team member, so it's trustworthy enough for now. But the Electrum servers are a big point of trust at the moment. Going forward we need to setup our own Electrum servers which we know to be trustworthy, as well as give users the option to connect to their own servers if they want to eliminate any trust in us.

sindresorhus: // @cipig

ptyx11: The server is ran and maintained by ChainZilla. We have it as a backup in case cipi's electrum network goes out.

PR #413, opened 2018-07-10
Fix quotes

lukechilds:
@sindresorhus @kevva can we merge this without messing up translations on Crowdin?

### DIFF
--- a/app/locales/en-US/dashboard.json
+++ b/app/locales/en-US/dashboard.json
@@ -1,7 +1,7 @@
 {
 	"intro": {
 		"title": "✨ Thanks for Helping Us Test HyperDEX!",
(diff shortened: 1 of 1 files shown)


sindresorhus:
What does this fix?

sindresorhus:
> can we merge this without messing up translations on Crowdin?

I have a feeling it will mess it up, at least the word inside the quotes...

lukechilds:
>What does this fix?

Currently it literally shows the text `&qoute;as is&qoute;` instead of `"as is"`.

>I have a feeling it will mess it up, at least the word inside the quotes...

😕

kevva:
If we don't want to break any translation strings we could use `dangerouslySetInnerHTML` for this one so that the HTML entities gets rendered correctly.

sindresorhus:
I'm gonna test and see if we can do this in a non-breaking way without using `dangerouslySetInnerHTML`.

PR #416, opened 2018-07-11
Automatically fix stuck swaps

### LINKED ISSUE #163: Support kickstarting swaps
https://modulelab.slack.com/archives/C8ADPSWTB/p1525328006000031

```
Sindre Sorhus [1:13 PM]
@Luke Childs https://github.com/jl777/SuperNET/commit/8725909b3c12f8df17de0b075422ad3676cc5195 

Have you thought about how the kickstart thing will work? I assume it's just a temporary workaround until mm does this automatically?

Luke Childs [1:14 PM]
no we’ll need kickstart long term
so if they quit the app with pending trades
or in process trades
then restart
we need to filter all the in progress trades and tell mm to resume them
I think it will be a fair bit of work to integrate
mm will most likely give us no helpful feedback in the event that:
- the trade doesn’t exist 
- exists but can’t be restarted
- was already automatically restarted
- is in the progress of being rolled back (edited)
we need to handle all those cases
and we probably won’t get enough information to
and we also need to re-attach event listeners so we get socket updates on restarted trades
if mm auto restarts them without telling us we’ll miss them
so many things that can go wrong
this is exactly why we shouldn’t have two separate swap DB instances…
😞
For v2 we need to push for a built in swap DB in mm that actually works and scales
so we have no internal copy
we just query mm as the single source of truth
They could use sqlite or LevelDB or something light weight
```

lukechilds:
Resolves #163 

This PR resolves the issue of users getting swaps that are stuck mid swap (e.g in `swap 3/5` state).

If a swap fails due to connectivity issues or closing the app the swap would become "stuck". Although the funds were cryptographically safe and reclaimable via the atomic swap protocol, `marketmaker` wouldn't actually reclaim the funds automatically without custom intervention via the console or CLI.

We are now manually looping over old swaps every 15 mins that are in a stuck state and "kickstarting" them to get `marketmaker` to recheck the status.

There are quite a lot of changes required to the current code base to get this working. I will try and break down the changes here and explain why they are necessary.

**Get updates on stuck swaps** https://github.com/hyperdexapp/hyperdex/pull/416/commits/1e7ceeea29c87658c7482767f09635375061cb00 https://github.com/hyperdexapp/hyperdex/pull/416/commits/5eaedd9dab0585a8836ff4e0bcdfc22739cf4ab0

We were previously listening for updates over the socket for a swap by adding a new event listener for each swap that listens out for messages containing it's UUID. This was not a great approach and resulted in a few crazy hacks to avoid memory leaks whilst also correctly handling quirky `marketmaker` behaviour where a swap would be reported as failed and the event listener was removed, only to later send us a completion message.

Now we only attach a single event listener and check all messages against a list of all known swap UUIDs. This means we can guarantee we will always catch updates to swaps, if they were started in a previous session and even if they are received in a strange order, e.g failed then complete. It's also much simpler code and avoids risks of memory leaks having a large amount of event listeners.

**Force kickstart stuck swaps** https://github.com/hyperdexapp/hyperdex/pull/416/commits/2e89499a46f768587ba98b41acc27d50f001e0e1 https://github.com/hyperdexapp/hyperdex/pull/416/commits/6514951a9634a24ba0455b3ffceab593773fb945

This is supposed to be handled automatically by `marketmaker`, but I've never personally seen it actually work. What we now do is loops over all swaps that are currently swapping and were started over 4 hours ago, then call `marketmaker`'s `kickstart` method. This forces it to check the current state of the swap and if something has gone wrong or Bob hasn't responded it will attempt to reclaim the funds from the timelocked address. This is why we only check if the swap is over 4 hours old, because the timelocks last 4 hours. We do this instantly on startup and then check again every 15 minutes.

On the first startup check, we check all pending swaps, even if they weren't started over 4 hours ago, this is because there are some scenarios where a swap can be reverted sooner.

This means you don't need to keep HyperDEX (and `marketmaker` running). You can close the app, open it the next day, and it will instantly reclaim your stuck swap if possible.

**Recalculate the state of a completed swap** https://github.com/hyperdexapp/hyperdex/pull/416/commits/fc1caab826820055f37fd2b6e1edc06691797c51 https://github.com/hyperdexapp/hyperdex/pull/416/commits/45b5b4910a0534bad2ea7b5e36bff7de6ea59174

Due to the fact that HyperDEX (and `marketmaker`) might not be running for the entire swap, we can't rely on the WebSocket message history to provide an accurate overview of what happened during the swap. The completed message we get at the end of every normal swap and "kickstarted" stuck swaps contains all the information we need to rebuild the full history of what happened.

So we now have two ways of building a formatted swap object. The socket messages are used while a swap is in progress. Once a swap is completed, the final completed socket message is used to reconstruct the entire swap transaction history and state which we know contains all data. The resulting structure of the swap object follows the same pattern in both methods so all existing UI code works no matter how the swap data was calculated.

This means if a swap is completed it should never have missing tx data, it should always list the entire chain of TXs, even if it wasn't online the whole time. It also means we correctly handle weird states like when the swap doesn't go to plan and we reclaim our original payment or claim `bobdeposit` instead of `bobpayment`. It also abstracts out quite a few quirks in `marketmaker` responses.

**Showing kickstarted swaps TXs in the swap modal** https://github.com/hyperdexapp/hyperdex/pull/416/commits/1a95811756701e33c139d2bb7cb3abdef431527c https://github.com/hyperdexapp/hyperdex/pull/416/commits/0e55952d967f4cd0afe2363e8fd678a66c8c70da

Kickstarted swaps don't always have the same transactions exposed as our current code expects which breaks the modal layout. I've modified the transactions array to not allow duplicate transactions (sometimes `marketmaker` sends multiple messages for the same TX which we were pushing into the transactions array). This means while the swap is in progress we can rely on just looping over the transaction array for the current data are less the 5 items fill them in with the placeholders.

Once a swap is completed however, there may not be 5 transactions due to the reversion and we don't want to show the placeholder boxes as those transactions are no longer expected to occur. However the reversion transactions are handled by the completed message and added to the transactions array, so just looping over that means they will show up, we just skip showing the placeholder boxes.

---

There's also a few other bits handling weird mm responses and adding extra data for stuck swaps, but they're fairly self explanatory.

To show how successfully resumed swaps look in the UI:

**A swap that was stuck, but then resumed in time to proceed as normal**

By either a reboot, losing network connectivity, accidentally closing the app etc.

This swap appears just like a normal swap that completed without issue:

screen shot 2018-07-12 at 3 21 59 pm

screen shot 2018-07-12 at 3 22 11 pm

*Notice the final TX is `alicespend`, the expected TX for a successful swap*

**A swap that was stuck due to Bob not sending `bobpayment`**

If Bob doesn't send `bobpayment` you need to wait four hours for the timelock on `bobdeposit` to expire. The we can claim `bobdeposit` with an `aliceclaim` TX. `bobdepost` is 12.5% more than the amount we were supposed to get. We get to keep this to punish Bob for being a dick. (In a normal trade Bob would reclaim his own `bobdeposit` after we spend `bobpayment` with `alicespend`).

You don't actually need to keep the app running for four hours though, you could go to bed and run the app in the morning and it will reclaim everything.

This swap appears just like a normal swap in the swap list. However in the modal it will show only four TXs in the TX chain and you'll notice that it shows a much higher percentage message than normal due to our 12.5% bonus:

screen shot 2018-07-12 at 3 24 10 pm

screen shot 2018-07-12 at 3 23 52 pm

*Notice the final TX is `aliceclaim` and is higher than what we asked for*

**A swap that was stuck due to Bob not sending any payments**

If Bob doesn't send any payments (`bobdeposit` or `bobpayment`) we can't really do much about it. We can reclaim our `alicepayment` funds but we lose our `dexfee` payment.

Although `marketmaker` reports the status as successfully completed (which kind of makes sense in terms of the atomic swap protocol), I'm mapping this to a failed swap in our logic because if the user doesn't get their desired funds, it's failed from their perspective.

So it will show in the swap list as reverted with the same styles as a failed/unmatched swap. The received value will be set as 0 and you will see the reclaim TX as the final TX in the chain. There is also a relatively vague message explaining what went wrong. (We really don't know what went wrong, it could have been many things):

screen shot 2018-07-12 at 3 24 21 pm

screen shot 2018-07-12 at 3 24 29 pm

*Notice the final TX is `alicereclaim` which just claims back our original funds*

---

As far as I'm aware, this correctly handles all stuck states and displays them correctly in the UI. You are guaranteed to either:

- Get the amount you were supposed to
- Get the amount you were supposed to + 12.5%
- Reclaim your original payment (but lose your `dexfee`)

### DIFF
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -38,10 +38,9 @@ class ExchangeContainer extends SuperContainer {
 			}
 		});
 
-		const FIFTEEN_MINUTES = 1000 * 60 * 15;
-		fireEvery(async () => {
+		fireEvery({minutes: 15}, async () => {
 			await this.kickstartStuckSwaps();
-		}, FIFTEEN_MINUTES);
+		});
 	}
 
 	constructor() {
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -40,8 +40,7 @@ class ExchangeContainer extends SuperContainer {
 
 		const ONE_HOUR = 1000 * 60 * 60;
 		fireEvery(async () => {
-			// Disable this while we test swap incentive flaw
-			// await this.kickstartStuckSwaps();
+			await this.kickstartStuckSwaps();
 		}, ONE_HOUR);
 	}
 
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -40,7 +40,8 @@ class ExchangeContainer extends SuperContainer {
 
 		const ONE_HOUR = 1000 * 60 * 60;
 		fireEvery(async () => {
-			await this.kickstartStuckSwaps();
+			// Disable this while we test swap incentive flaw
+			// await this.kickstartStuckSwaps();
 		}, ONE_HOUR);
 	}
 
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -1,6 +1,7 @@
 /* eslint-disable react/no-access-state-in-setstate */
 import {is, api, activeWindow, appLaunchTimestamp} from 'electron-util';
 import _ from 'lodash';
+import {isPast, addHours} from 'date-fns';
 import SuperContainer from 'containers/SuperContainer';
 import appContainer from 'containers/App';
 import {translate} from '../translate';
@@ -25,8 +26,8 @@ class ExchangeContainer extends SuperContainer {
 		};
 	}
 
-	componentDidInitialMount() {
-		this.setSwapHistory();
+	async componentDidInitialMount() {
+		await this.setSwapHistory();
 		appContainer.swapDB.on('change', this.setSwapHistory);
 		appContainer.api.socket.on('message', message => {
 			const uuids = this.state.swapHistory.map(swap => swap.uuid);
@@ -35,6 +36,11 @@ class ExchangeContainer extends SuperContainer {
 				appContainer.swapDB.updateSwapData(message);
 			}
 		});
+
+		const ONE_HOUR = 1000 * 60 * 60;
+		fireEvery(async () => {
+			await this.kickstartStuckSwaps();
+		}, ONE_HOUR);
 	}
 
 	constructor() {
@@ -54,6 +60,17 @@ class ExchangeContainer extends SuperContainer {
 				this.setQuoteCurrency(newQuoteCurrency);
 			}
 		});
+	}
+
+	async kickstartStuckSwaps() {
+		this.state.swapHistory
+			.filter(swap => {
+				return swap.status === 'swapping' && isPast(addHours(swap.timeStarted, 4));
+			})
+			.forEach(async swap => {
+				const {requestId, quoteId} = swap;
+				const result = await appContainer.api.kickstart({requestId, quoteId});
+			});
 	}
 
 	setSwapHistory = async () => {
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -38,10 +38,10 @@ class ExchangeContainer extends SuperContainer {
 			}
 		});
 
-		const ONE_HOUR = 1000 * 60 * 60;
+		const FIFTEEN_MINUTES = 1000 * 60 * 15;
 		fireEvery(async () => {
 			await this.kickstartStuckSwaps();
-		}, ONE_HOUR);
+		}, FIFTEEN_MINUTES);
 	}
 
 	constructor() {
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -72,7 +72,7 @@ class ExchangeContainer extends SuperContainer {
 			))
 			.forEach(async swap => {
 				const {requestId, quoteId} = swap;
-				const result = await appContainer.api.kickstart({requestId, quoteId});
+				await appContainer.api.kickstart({requestId, quoteId});
 			});
 
 		if (!doneInitialKickstart) {
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -31,6 +31,7 @@ class ExchangeContainer extends SuperContainer {
 		appContainer.api.socket.on('message', message => {
 			const uuids = this.state.swapHistory.map(swap => swap.uuid);
 			if (uuids.includes(message.uuid)) {
+				console.log(message.method, message);
 				appContainer.swapDB.updateSwapData(message);
 			}
 		});
--- a/app/renderer/containers/Exchange.js
+++ b/app/renderer/containers/Exchange.js
@@ -23,6 +23,7 @@ class ExchangeContainer extends SuperContainer {
 				askdepth: 0,
 			},
 			isSendingOrder: false,
+			doneInitialKickstart: false,
 		};
 	}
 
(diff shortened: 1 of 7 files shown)


lukechilds:
Also, I'm aware the `_formatSwap()` function has grown a little bit out of control. It's now doing way more than we ever intended it to. It's handling multiple different ways of reading swap data and abstracting all of `marketmaker`'s weirdness away to try and expose a single uniformly formatted swap object.

This function should probably be refactored into it's own file with more readable methods at some point.

sindresorhus:
> This function should probably be refactored into it's own file with more readable methods at some point.

Can you add a `TODO` comment in the code so we don't forget?

app/renderer/containers/Exchange.js
```diff
@@ -22,12 +23,25 @@ class ExchangeContainer extends SuperContainer {
...
+		appContainer.api.socket.on('message', message => {
+			const uuids = this.state.swapHistory.map(swap => swap.uuid);
+			if (uuids.includes(message.uuid)) {
+				console.log(message.method, message);
+				appContainer.swapDB.updateSwapData(message);
+			}
+		});
+
+		const FIFTEEN_MINUTES = 1000 * 60 * 15;
+		fireEvery(async () => {
```
sindresorhus: `fireEvery({minutes: 15}, async () => {`

app/renderer/swap-db.js
```diff
@@ -144,43 +147,144 @@ class SwapDB {
...
+					const [alicespend, bobspend, bobpayment, alicepayment, bobdeposit, otherfee, myfee, bobrefund, bobreclaim, alicereclaim, aliceclaim] = message.values;
+					return {alicespend, bobspend, bobpayment, alicepayment, bobdeposit, otherfee, myfee, bobrefund, bobreclaim, alicereclaim, aliceclaim};
+				})();
+
+				swap.transactions = [];
+				if (message.sentflags.includes('myfee')) {
+					swap.transactions.push({
+						stage: 'myfee',
+						coin: message.alice,
+						txid: undefined, // This is currently mising from marketmaker responses
```
sindresorhus: Can you link to a marketmaker issue?

app/renderer/swap-db.js
```diff
@@ -144,43 +147,144 @@ class SwapDB {
...
+					});
+				}
+
+				// This is the final tx in the case that bob doesn't send bobpayment.
+				// We can claim bobdeposit which gives us a 12.5% bonus to punish bob.
+				if (message.sentflags.includes('aliceclaim')) {
+					// There is a bug in marketmaker where it doesn't always correctly report
+					// the values. If aliceclaim is 0 we fallback to bobdeposit as it should
+					// be very close (same amount minus our claim txfee)
+					// https://github.com/jl777/SuperNET/issues/920
```
sindresorhus: This seems to be fixed now? https://github.com/jl777/SuperNET/issues/920#issuecomment-403392467

lukechilds: Yep, should be, yet to check that.

app/renderer/swap-db.js
lukechilds: There is no GH issue, but I mentioned it to James and he said he'd fixed it. Will test and create an issue if it isn't resolved.

app/renderer/swap-db.js
lukechilds: Doesn't appear to be resolved https://github.com/jl777/SuperNET/issues/920#issuecomment-404815003

sindresorhus:
Oh yeah! Feels good 🙌

d8ee7104-273c-11e5-9676-43241f2fc082

PR #426, opened 2018-07-17
Enable ERC20 tokens

lukechilds:
This enables the ERC20 tokens as they're now fully working.

There are barely any Bobs so it's hard to actually test. I've setup a Unicoin Bob for testing but it costs me ETH every time you make a trade so be gentle.

The remaining fixes:
- Update mm: https://github.com/atomiclabs/hyperdex/commit/e3538f87c68ecd18040def3c511fc7cc8b191da8
- ~~Handle ETOMIC TXIDs: https://github.com/atomiclabs/hyperdex/commit/f26f256391d99dff7409b905732b26d7f738f471~~ No longer needed

are already in master.


lukechilds:
Actually, just noticed there is a bug, we show the ETOMIC amounts which are always 1, not the actual ETH/ERC20 amounts.

We will likely need changes to mm to resolve this.

lukechilds:
These are the final (🤞) changes we need in `marketmaker` for full ETH/ERC20 support: https://github.com/artemii235/SuperNET/issues/104

lukechilds:
This is resolved now thanks to @artemii235's great work. Pretty much all the transaction chain building logic we had in our app has been implemented inside `marketmaker` now so we get a neat little array in the completion method of all the transaction details.

Only downside is this is a complete breaking change. Not just for people with ETH trades but for all trade history of any pairs.

We _could_ add backwards compatibility but I'd really prefer not to as this has allowed us to strip a lot of complexity out of the swap message handling. We'd have to add all that logic back in wrapped in a load of if statements for backwards compat.

I would strongly suggest we ship this as is and let people know that they should wipe their swap DB before updating to the next release. What are your thoughts @atomiclabs/core?

sindresorhus:
> I would strongly suggest we ship this as is and let people know that they should wipe their swap DB before updating to the next release.

👍

sindresorhus:
Oh yeah!

PR #431, opened 2018-07-18
Add hotkey to restart HyperDEX in debug mode

sindresorhus:
I first implemented this as a shortcut (Control+Shift) you hold down during app launch to activate debug-mode. Well, turns out you can't do shortcuts without a non-modifier ([#9900](https://github.com/electron/electron/issues/9900)) in Electron. Then there's the issue that Electron doesn't allow you to subscribe to shortcuts until `app.isReady()` ([#13716](https://github.com/electron/electron/issues/13716)), which is too late for us as we've already done the debug-mode check by then. Then I ended doing it so that you can press Control+Shift+Alt+D after the app has launched. Of course, our journey doesn't end there. Turns out there's an [`electron-builder` issue](https://github.com/electron-userland/electron-builder/issues/1727) which prevents the AppImage (Linux) from being relaunched. The ability to relaunch is important as we relaunch with the `--debug` flag to indicate debug-mode. I then considered moving the debug state from the `--debug` flag to saving it to our config and reading it back on launch and clear it out if it exists.

Before I waste more time on this. Do you think it should be a keyboard shortcut or should we maybe put it in the `Help` menu as `Debug Mode` that can be toggled?

// @lukechilds @kevva

### DIFF
--- a/app/util-common.js
+++ b/app/util-common.js
@@ -4,7 +4,7 @@ const {api} = require('electron-util');
 const isNightlyBuild = api.app.getName() === 'HyperDEX Nightly';
 
 /// TODO: Change this before the official launch
-/// const isDevelopment = is.development || isNightlyBuild;
+/// const isDevelopment = process.argv.slice(1).includes('--debug') || is.development || isNightlyBuild;
 const isDevelopment = true;
 
 module.exports = {
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -35,3 +35,7 @@ $ ./signedchecksum
 ```
 
 It will create a file called `SHASUMS256.txt.asc` that you need to upload to the GitHub release manually.
+
+## Debug-mode in production
+
+Press <kbd>Control</kbd> <kbd>Shift</kbd> <kbd>Alt</kbd> <kbd>D</kbd> to relaunch HyperDEX in debug-mode. You can then access debug features in the Debug menu.
--- a/app/index.js
+++ b/app/index.js
@@ -28,7 +28,7 @@ try {
 	require('electron-reloader')(module, {watchRenderer: false});
 } catch (_) {}
 
-const {app, session} = electron;
+const {app, session, globalShortcut} = electron;
 
 app.setAppUserModelId('com.lukechilds.hyperdex');
 
@@ -145,7 +145,21 @@ if (!is.development) {
 	`);
 }
 
+const registerDebugHotkey = () => {
+	const debugShortcut = 'Control+Shift+Alt+D';
+
+	globalShortcut.register(debugShortcut, () => {
+		globalShortcut.unregister(debugShortcut);
+		app.relaunch({
+			args: process.argv.slice(1).concat(['--debug']),
+		});
+		app.quit();
+	});
+};
+
 app.on('ready', () => {
+	registerDebugHotkey();
+
 	session.defaultSession.setPermissionRequestHandler((webContents, permission, callback) => {
 		if (permission === 'notifications') {
 			callback(true);


lukechilds:
If it's gonna cause headaches I think having it as a menu item is fine.

sindresorhus:
@lukechilds It's not gonna cause headaches. I just want us to decide which way to do it before I spend time implementing it. The benefit of a menu item is ease of use, but that's also its downside. It means users could discover it and do stuff they shouldn't.

sindresorhus:
I added it to the Help menu. In development and nightly builds, the value is ignored and not changeable. In production, you can click `Debug Mode` and the HyperDEX will restart in debug mode.

app/util-common.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
 const {api} = require('electron-util');
+/// const config = require('./config');
 
 const isNightlyBuild = api.app.getName() === 'HyperDEX Nightly';
 
 /// TODO: Change this before the official launch
-/// const isDevelopment = is.development || isNightlyBuild;
+/// const isDevelopment = config.get('isDebugMode') || is.development || isNightlyBuild;
 const isDevelopment = true;
```
sindresorhus: To test it out, you need to comment this and uncomment the above, the run `yarn pack` and run the build in `dist/mac/HyperDEX.app`.

lukechilds: >run yarn pack and run the build in dist/mac/HyperDEX.app

Did you mean `yarn dist` here? `yarn pack` only works for me in `hyperdex/app` and gives me an archive.

lukechilds: If I run `yarn dist` I get the file at `dist/mac/HyperDEX.app` but running it just gives me:

screen shot 2018-07-25 at 4 26 54 pm

lukechilds: Ah, looks like we need to import `is` on line 2 too.

```js
const {api, is} = require('electron-util');
```

🤦‍♂️

lukechilds:
These are the full changes we need to work and make sure `is` is pulled in:

```patch
diff --git a/app/util-common.js b/app/util-common.js
index ff92701e..7fba61fb 100644
--- a/app/util-common.js
+++ b/app/util-common.js
@@ -1,12 +1,12 @@
 'use strict';
-const {api} = require('electron-util');
-/// const config = require('./config');
+const {api, is} = require('electron-util');
+const config = require('./config');
 
 const isNightlyBuild = api.app.getName() === 'HyperDEX Nightly';
 
 /// TODO: Change this before the official launch
-/// const isDevelopment = config.get('isDebugMode') || is.development || isNightlyBuild;
-const isDevelopment = true;
+const isDevelopment = config.get('isDebugMode') || is.development || isNightlyBuild;
+// const isDevelopment = true;
 
 module.exports = {
 	isNightlyBuild,
```

lukechilds:
What do you think about enabling it in dev mode too so we can quickly toggle to production mode for testing stuff?

sindresorhus:
> What do you think about enabling it in dev mode too so we can quickly toggle to production mode for testing stuff?

Testing what though? You don't really get a real production environment by turning of debug mode when developing. I didn't want us to be fooled into thinking so. Real production builds with optimizations in webpack, it builds the files differently, moves dependencies differently, puts everything into a container file (ASAR), etc.

sindresorhus:
> These are the full changes we need to work and make sure is is pulled in:

Yeah, sorry, I forgot to mention having to add `is`.

lukechilds:
>You don't really get a real production environment by turning of debug mode when developing.

Oh ok, no point then.

PR #434, opened 2018-07-19
Render numbers with decimal instead of exponential notation in `Input`

### LINKED ISSUE #318: The decimal point is displayed as text
image

Step :
1. Input purchase price is 0.00250835BCH.
2. Click the Max button to enter the maximum quantity.
3. Check Total (BCH):

- I think it should be displayed as 0.0000002 instead of 2e-7.

HyperDEX 0.1.0-alpha.1
Electron 2.0.0
win32 10.0.17134
Locale: ko

kevva:
Fixes #318.


app/renderer/components/Input.js
```diff
@@ -124,9 +128,18 @@ class Input extends React.Component {
...
 		if (onlyNumeric) {
+			if (this._isExponentialNotation(value)) {
+				value = Number.parseFloat(value).toLocaleString('fullwide', {
+					maximumFractionDigits: 20, // 20 is the maximum, we truncate below instead
+				});
+			}
+
 			if (this._shouldTruncateFractions(value)) {
-				value = Number.parseFloat(value).toFixed(fractionalDigits);
+				value = Number.parseFloat(value).toLocaleString('fullwide', {
```
kevva: Not a fan of running the same stuff twice, but it's late and my brain can't think of a better way since we need to convert it to a decimal string before we know if we should truncate it.

sindresorhus: How about this?

```js
const shouldTruncate = this._shouldTruncateFractions(value);
if (this._isExponentialNotation(value) || shouldTruncate) {
	value = Number.parseFloat(value).toLocaleString('fullwide', {
		maximumFractionDigits: shouldTruncate ? fractionalDigits : 20,
	});
}
```

sindresorhus:
Can you add a [custom proptype](https://stackoverflow.com/questions/41821181/react-proptypes-number-float-with-min-and-max) to ensure the `fractionalDigits` prop is not more than 20?

sindresorhus:
If I put `1` in Price and `999999999999999999999` in Total, Amount ends up being `1,000,000,000,000,000,000,000`. It should never have `,`, and now I can't edit the Amount field.

app/renderer/components/Input.js
kevva: If it's exponential it wouldn't know if it should truncate it since `_shouldTruncateFractions` only works on decimal numbers. That's why we need to convert it before.

kevva:
> If I put 1 in Price and 999999999999999999999 in Total, Amount ends up being 1,000,000,000,000,000,000,000. It should never have ,, and now I can't edit the Amount field.

Hm, that's weird. `fullwide` should make it so it doesn't use decimal separators.

app/renderer/components/Input.js
sindresorhus: Hmm, ok. Then the only simplification I can think of is:

```js
const truncateFractions = maximumFractionDigits => Number.parseFloat(value).toLocaleString('fullwide', {maximumFractionDigits});
```

But not sure it's worth the abstraction.

kevva: I think it is.

kevva:
Never mind, didn't see it was a positive number. `useGrouping: false` fixed it.

package.json
```diff
@@ -32,6 +32,7 @@
 		"pouchdb-browser": "^7.0.0",
 		"pouchdb-find": "^7.0.0",
 		"prop-types": "^15.6.2",
+		"prop-types-range": "^0.0.0",
```
kevva: Lol, forgot to bump it.

PR #436, opened 2018-07-19
Add success/fail color to the progress bar in the swap modal

sindresorhus:
The progress bar turns green when completed and red when there's an error.

screen shot 2018-07-19 at 17 25 50

screen shot 2018-07-19 at 17 25 35


lukechilds:
Love it.

Issue #177, opened 2018-05-07
Swap modal tweaks


sindresorhus:
It just needs a few more tweaks. I'll be working on this today.

sindresorhus:
Some minor things still left, but not a priority:

- [ ] Don't show the detailed info by default, but make it toggleable. The choice should be persisted.
- [ ] Experiment with having the progress bar at the top of the modal.

PR #437, opened 2018-07-20
Move the swap modal progressbar to the top of the modal

sindresorhus:
Per request in #177. I'm still undecided on this.

@lukechilds @kevva Thoughts?

#### Before

screen shot 2018-07-20 at 18 37 57

#### After

screen shot 2018-07-20 at 18 27 58
screen shot 2018-07-20 at 18 27 49

### DIFF
--- a/app/renderer/components/Modal.scss
+++ b/app/renderer/components/Modal.scss
@@ -129,8 +129,10 @@
 			}
 
 			h1 {
+				width: 100%;
 				font-size: 17px;
 				margin: 0;
+				margin-right: 20px;
 				padding: 0;
 				line-height: 1;
 			}
--- a/app/renderer/components/SwapDetails.js
+++ b/app/renderer/components/SwapDetails.js
@@ -109,11 +109,19 @@ class SwapDetails extends React.Component {
 
 		const overview = getOverview(swap);
 
+		const titleComponent = (
+			<div className="title">
+				<div>{title(swap.statusFormatted)}</div>
+				<div className="title__main">{baseCurrency}/{quoteCurrency} {t(`details.${swap.orderType}`)} {t('details.order')}</div>
+				<div>{formatDate(swap.timeStarted, 'HH:mm DD/MM/YY')}</div>
+			</div>
+		);
+
 		return (
 			<div className="modal-wrapper">
 				<Modal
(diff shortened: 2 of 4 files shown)


lukechilds:
I definitely prefer it visually, when the progress bar is at the top.

However it's nice in the previous version that the progress bar and "swap n/5" text are right next to each other. The progress text looks a little out of place on the current version.

One potential solution would be to keep the progress bar at the top, have the status text in the modal body as "Swapping" and add "Swap n/5" to the title, so it's visually associated with the progress bar again.

Thoughts?

sindresorhus:
@lukechilds How about this? In the title, have the status on the left side, the type of order on the middle and the date on the right side? (And make the left/right darker or smaller to highlight the center one)

screen shot 2018-07-24 at 04 04 30

kevva:
> And make the left/right darker or smaller to highlight the center one (Not done in the screenshot).

Yes. Looks pretty crowded in the screenshot. Not sure how obvious it'd be that the   top left status (like 3/5 swaps) ties into the progress bar. The `Completed` status looks pretty nice to me though. Definitely like it better at the top.

lukechilds:
@sindresorhus I really like that.

Assuming we'll use the formatted status in the top left so it'll be "Swap 2/5" not "Swapping"?

sindresorhus:
This is not yet ready to be merged. I have to implement what was decided on here.

sindresorhus:
The screenshot was just a quick mockup :p

lukechilds:
Woops, got a bit trigger happy on the review button.

sindresorhus:
How it looks now:

screen shot 2018-07-30 at 21 33 41
screen shot 2018-07-30 at 21 33 48

Thoughts?

lukechilds:
I like it.

Shouldn't there be a red bar at the top of the failed swap?

sindresorhus:
> Shouldn't there be a red bar at the top of the failed swap?

There is no red bar as there is no progress. It failed at the pre-step. Do you think it would make sense to still force a 100% red progress bar?

lukechilds:
Ah of course, I think so yeah, just for consistency.

sindresorhus:
Done

PR #443, opened 2018-07-23
Fix duplicate labels on the TimeSeriesChart x-axis

### LINKED ISSUE #311: Duplicated labels for the "1 Year" scale in the currency price history graph
screen shot 2018-05-25 at 14 06 15

Maybe some help here: https://github.com/recharts/recharts/issues/1137

benv2:
Did my best to fix #311

I'm pretty new to open source so I'm not entirely sure if I'm doing this right.

### DIFF
--- a/app/renderer/components/TimeSeriesChart.js
+++ b/app/renderer/components/TimeSeriesChart.js
@@ -25,25 +25,18 @@ const resolutionToLabelFormat = new Map([
 	['all', 'MMMM YYYY'],
 ]);
 
-const now = new Date();
-const makeTicks = (count, unit) => {
-	const subFunction = {
-		month: (date, index) => subMonths(date, index),
-		day: (date, index) => subDays(date, index),
-		hour: (date, index) => subHours(date, index),
-		minute: (date, index) => subMinutes(date, index),
-	};
-
-	return [...new Array(count).keys()].map(index => subFunction[unit](now, index)).reverse();
+const makeTicks = (count, fn) => {
+	const now = new Date();
+	return [...new Array(count).keys()].map(index => fn(now, index)).reverse();
 };
 
 const getTicks = resolution => {
 	const ticks = {
-		year: makeTicks(13, 'month'),
-		month: makeTicks(getDaysInMonth(now) + 1, 'day'),
-		week: makeTicks(8, 'day'),
-		day: makeTicks(25, 'hour'),
-		hour: makeTicks(61, 'minute'),
+		year: makeTicks(13, subMonths),
+		month: makeTicks(getDaysInMonth(new Date()) + 1, subDays),
+		week: makeTicks(8, subDays),
+		day: makeTicks(25, subHours),
+		hour: makeTicks(61, subMinutes),
 	};
 	return ticks[resolution];
 };
--- a/app/renderer/components/TimeSeriesChart.js
+++ b/app/renderer/components/TimeSeriesChart.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import _ from 'lodash';
 import {ResponsiveContainer, AreaChart, Area, XAxis, YAxis, Tooltip} from 'recharts';
-import {format as formatDate} from 'date-fns';
+import {format as formatDate, subMonths, subDays, subHours, subMinutes, getDaysInMonth} from 'date-fns';
 import {formatCurrency} from '../util';
 import './TimeSeriesChart.scss';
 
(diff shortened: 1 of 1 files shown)


app/renderer/components/TimeSeriesChart.js
```diff
@@ -36,8 +36,33 @@ class TimeSeriesChart extends React.Component {
...
 		const labelFormat = resolutionToLabelFormat.get(resolution);
 
+		const getTicks = (resolution) => {
+			const ticks = {
+				"all": undefined,
+				"year": [...Array(13).keys()].map(index => subMonths(new Date(), index)).reverse(),
+				"month": [...Array(getDaysInMonth(new Date()) + 1).keys()].map(index => subDays(new Date(), index)).reverse(),
+				"week": [...Array(8).keys()].map(index => subDays(new Date(), index)).reverse(),
+				"day": [...Array(25).keys()].map(index => subHours(new Date(), index)).reverse(),
+				"hour": [...Array(61).keys()].map(index => subMinutes(new Date(), index)).reverse(),
```
kevva: Are these numbers arbitrary or where do they come from?

app/renderer/components/TimeSeriesChart.js
```diff
@@ -36,8 +36,33 @@ class TimeSeriesChart extends React.Component {
 			resolution,
 		} = this.props;
 
+		console.log(data);
```
kevva: Remove this.

sindresorhus:
Thanks for working on this, @benv2 😃👍

sindresorhus:
Make sure you run `npm test` locally and fix any lint violations.

app/renderer/components/TimeSeriesChart.js
benv2: Maybe I should have explained this.
To prevent duplicate labels I set the ticks manually.
The hour chart should have a maximum of 61 ticks, meaning 1 for every minute in an hour plus one more so the first and last ticks have the same time.
Likewise, the day chart should have 25 ticks since there are 24 hours in a day + 1 hour for the first tick.
It works analogously for the week and the month chart.

benv2:
I fixed the linting violations and improved the code a bit.
Also removed the console.log kevva mentioned.

app/renderer/components/TimeSeriesChart.js
```diff
@@ -38,6 +38,17 @@ class TimeSeriesChart extends React.Component {
 
 		const labelFormat = resolutionToLabelFormat.get(resolution);
 
+		const getTicks = resolution => {
+			const ticks = {
+				year: [...new Array(13).keys()].map(index => subMonths(new Date(), index)).reverse(),
+				month: [...new Array(getDaysInMonth(new Date()) + 1).keys()].map(index => subDays(new Date(), index)).reverse(),
+				week: [...new Array(8).keys()].map(index => subDays(new Date(), index)).reverse(),
+				day: [...new Array(25).keys()].map(index => subHours(new Date(), index)).reverse(),
+				hour: [...new Array(61).keys()].map(index => subMinutes(new Date(), index)).reverse(),
```
sindresorhus: Nitpick, but it's wasteful to create so many date objects. I would define `const now = new Date();` above `const ticks` and use that.

You could also abstract some of the logic into a helper function:

```js
const makeTicks = (count, fn) => {
	return [...new Array(count).keys()].map(index => fn(new Date(), index)).reverse();
};
```

app/renderer/components/TimeSeriesChart.js
```diff
@@ -38,6 +38,17 @@ class TimeSeriesChart extends React.Component {
 
 		const labelFormat = resolutionToLabelFormat.get(resolution);
 
+		const getTicks = resolution => {
```
sindresorhus: You can move this to the top-level, below the `resolutionToLabelFormat` function, since it doesn't depend on any internal state.

benv2:
As you suggested I changed the multiple new Dates to a const.
I also introduced a "helper" function but made it a bit more verbose than your example.

This this possible:
https://github.com/benv2/hyperdex/blob/b258c075d0bcfb6e9f92a573f92f3920b2203658/app/renderer/components/TimeSeriesChart.js#L40-L47

app/renderer/components/TimeSeriesChart.js
```diff
@@ -25,6 +25,29 @@ const resolutionToLabelFormat = new Map([
...
 ]);
 
+const now = new Date();
+const makeTicks = (count, unit) => {
+	const subFunction = {
+		month: (date, index) => subMonths(date, index),
+		day: (date, index) => subDays(date, index),
+		hour: (date, index) => subHours(date, index),
+		minute: (date, index) => subMinutes(date, index),
+	};
```
sindresorhus: I don't really see the point of this. It just makes it more verbose. `makeTicks(13, 'month')` is not much clearer than `makeTicks(13, subMonths)`.

app/renderer/components/TimeSeriesChart.js
```diff
@@ -25,6 +25,29 @@ const resolutionToLabelFormat = new Map([
 	['all', 'MMMM YYYY'],
 ]);
 
+const now = new Date();
```
sindresorhus: This needs to be inside the `makeTicks` function as we need the date when `getTicks` called, not when the module is evaluated.

app/renderer/components/TimeSeriesChart.js
benv2: My bad. I forgot I could write it like that.
I passed '(date, index) => subDays(date, index)' instead of 'subDays' which made it look relatively messy.
Will fix this!

benv2:
I improved the code with your suggestions. 👍 

I noticed though that sometimes the chart has more data than necessary. This results in the first tick being offset. This is especially noticeable on the day chart.

bildschirmfoto 2018-07-23 um 21 57 34

sindresorhus:
This is looking great! We really appreciate the help, @benv2 🙌

sindresorhus:
> I noticed though that sometimes the chart has more data than necessary. This results in the first tick being offset. This is especially noticeable on the day chart.

Weird. I can't reproduce that. Maybe it depends on when in the day you check. If you still see this, would you be interested in looking into it?

sindresorhus:
> This is a bit off-topic but I was wondering if you still need help translating the app.
German is my mother tongue so maybe I could help with that.

Definitely, that would be great! We still have some untranslated strings in German and we expect more in the future. I've sent you an invite to our [Crowdin project](https://crowdin.com/project/hyperdex) (It's what we use to translate the app). Feel free to look through the existing translated strings too and see if you find any mistakes or improvements ;)

benv2:
@sindresorhus Cool. Will do! 👍

PR #444, opened 2018-07-24
Get rid of Sass variables

sindresorhus:
This converts the remaining Sass variables to React icons. And disallows using Sass variables in the future. We should prefer native CSS variables as they work live, which is needed for theming.


sindresorhus:
> I much prefer having the SVGs as React components.

Totally. The only reason I didn't do this initially is that I didn't realize React supported SVG natively.

PR #445, opened 2018-07-24
Limit the amount of swaps in the Exchange view

sindresorhus:
It's just naively slicing off what we need. This is not meant as a performance optimization, but rather to make it less noisy for the user. All swaps are available in the Trades view anyway.

### DIFF
--- a/app/locales/en-US/exchange.json
+++ b/app/locales/en-US/exchange.json
@@ -29,6 +29,7 @@
 	"quit": "Quit",
 	"swaps": {
 		"all": "All",
-		"title": "Recent Swaps"
+		"title": "Recent Swaps",
+		"viewAllSwaps": "View all swaps"
 	}
 }
--- a/app/locales/en-US/exchange.json
+++ b/app/locales/en-US/exchange.json
@@ -29,6 +29,6 @@
 	"quit": "Quit",
 	"swaps": {
 		"all": "All",
-		"title": "Swaps"
+		"title": "Recent Swaps"
 	}
 }
--- a/app/renderer/components/SwapList.js
+++ b/app/renderer/components/SwapList.js
@@ -2,12 +2,14 @@ import {api} from 'electron-util';
 import React from 'react';
 import {format as formatDate} from 'date-fns';
 import appContainer from 'containers/App';
+import tradesContainer from 'containers/Trades';
 import Empty from 'components/Empty';
+import Link from 'components/Link';
 import SwapDetails from 'components/SwapDetails';
 import {translate} from '../translate';
 import './SwapList.scss';
 
-const t = translate('swap');
+const t = translate(['swap', 'exchange']);
 
 // eslint-disable-next-line no-unused-vars
 class CancelButton extends React.Component {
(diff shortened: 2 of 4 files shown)


sindresorhus:
I just arbitrarily picked 50. Happy to revise that if someone prefers a different number.

kevva:
> I just arbitrarily picked 50. Happy to revise that if someone prefers a different number.

50 is okay. I'm wondering if we should add some sort of message that indicates that we're just showing the last 50 swaps and not all of them though?

sindresorhus:
@kevva How about renaming the title from `Swaps` to `Recent Swaps`?

kevva:
@sindresorhus, that'd be better I think. At least gives some explanation to why not all of them are listed.

lukechilds:
Is there any issue with showing all swaps?

Unless we're seeing React perf issues I'm not sure it makes sense to limit it. I often want to scroll pretty far back when I'm testing things to compare swaps (although this is more of a dev requirement, probs not something many users would be doing)

>How about renaming the title from Swaps to Recent Swaps?

👍

kevva:
> Is there any benefit to limiting the swaps? I'm not sure it's reducing noise as you only see the amount of swaps visible in the scroll view at a given time.

Most exchanges limit them in some way, either by using filters (like time) or by having a pagination. Maybe just limiting them to a certain number isn't the way to go... 

As a trader I think it's pretty useful to have the history easily available. But in other exchanges it's not just one click away like it is in HyperDEX where you can go to the Trades view at all times.

sindresorhus:
> Is there any benefit to limiting the swaps?

The more you have, the more sensitive the scrolling will be, and it will also affect React performance over some absurd number. We have to limit it to something.

How about we change the number to 100 then? We can have a lot there, just not thousands of swaps.

lukechilds:
>The more you have, the more sensitive the scrolling will be

That's a good point


>Why not just list the 50 most recent and then more upon scrolling to the bottom then?

Or a "Load More" button if it's going to cause jankyness injecting them dynamically. Unless we can pull it off super smooth I thinking auto injecting into scroll bars always looks a bit dodgy.

sindresorhus:
Why complicate it so much? It's just meant as a widget to see current and recent swaps. For history and more advanced filtering and searching, we have the Trades view.

lukechilds:
Ok, we should make it clear at the bottom that they should view the order history to view all trades.

Just to avoid confusion so they don't think trades are disappearing.

What about adding a final row that says "View all swaps" which when clicked jumps to the Trades view.

Or maybe just the "Recent Swaps" title does enough to convey that information.

sindresorhus:
> What about adding a final row that says "View all swaps" which when clicked jumps to the Trades view.

I was going to suggest that too. Let's do it.

lukechilds:
>I was going to suggest that too

Of course you were hun.

sindresorhus:
Added in https://github.com/atomiclabs/hyperdex/pull/445/commits/c2c3572d04e01b5dea9e70ee68138d1b76d5bef3

This is how it looks when you scroll to the end:

screen shot 2018-07-27 at 22 33 12

lukechilds:
Looks great 👌

PR #453, opened 2018-07-26
Link transactions to block explorer web pages

lukechilds:
Links the transaction blocks in the swap modal to the actual transaction pages on the block explorers.

If we ever link to a currency we don't have a block explorer URL for it will fallback to `https://www.google.com/search?q=${symbol} Transaction ${txid}`.

Also added a `isEtomic()` utility method to `supported-currencies.js`.

blockexplorerlinks

The hover styles are just the default link hover styles. I haven't changed them but I think they look pretty good as is.

Close-up:

hover effect


app/renderer/block-explorer.js
```diff
@@ -0,0 +1,185 @@
+// TODO(lukechilds): Extract this into an npm package when it's more mature
+import ow from 'ow';
+import {isEtomic} from '../marketmaker/supported-currencies';
+
+const explorers = {
```
sindresorhus: I would have used a `Map` here as it's more efficient but doesn't matter much.

lukechilds: I thought you might say that hehe.

I'm happy to change it. What are the benefits?

sindresorhus: Performance, but mostly because it's the correct type for this kind of key value store. I know the syntax is ugly though, so you could just create a Map from the Object at evaluation time.

Like:

```js
const explorers = {
	ANC: 'http://abe.darkgamex.ch:2751/tx/{txid}',
	ARC: 'http://explorer.arcticcoin.org/tx/{txid}',
};

const map = new Map(Object.entries(explorers));
```

lukechilds: Just FYI, I've written this with some future behaviour in mind.

Currently I've exposed the function as `blockExplorer.tx(symbol, txid)`. That's because I'm planning on exposing `blockExplorer.address(symbol, address)` in the future.

Which will likely mean the `explorers` schema will change to something like:

```js
const explorers = {
	ANC: {'http://abe.darkgamex.ch:2751/tx/{txid}',
	ARC: 'http://explorer.arcticcoin.org/tx/{txid}',
};
```

lukechilds: Actually, I guess it's still just a map of keys (symbols) to objects (explorer URLs) so a Map still makes sense, right?

sindresorhus: Yup. In a Map, the key and value can be any type.

lukechilds:
@sindresorhus better? https://github.com/atomiclabs/hyperdex/pull/453/commits/1a1665baaf2fba7a5dbddb05b56e31ec11228901

app/renderer/block-explorer.js
```diff
@@ -0,0 +1,185 @@
...
+	ow(txid, ow.string.label('txid'));
+
+	const explorer = explorers.get(isEtomic(symbol) ? 'ETH' : symbol);
+
+	// Fallback
+	if (!explorer) {
+		return `https://www.google.com/search?q=${symbol} Transaction ${txid}`;
+	}
+
+	const explorerUrl = explorer.replace('{txid}', txid);
```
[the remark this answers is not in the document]
sindresorhus: That's cleaner indeed, but then we would have to recreate the whole Map for each request to it, which might have a performance penalty.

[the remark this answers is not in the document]
lukechilds: I know `String#replace()` isn't the best way of doing this purely in JS.

I intentionally did it like that because I'm planning on extracting this out to a module where the explorers object will be an external JSON.

That way anyone can pull it from npm/GitHub, regardless of what programming language they use and easily parse the urls.

PR #462, opened 2018-08-03
Add some useful readme badges

sindresorhus:
I'm aware the Crowdin badge doesn't totally fit in. I've emailed them about that.

The chat badge would show how many people we have in our Discord channel. But we need one of the admine to turn it on:

> In discord Server Server Settings -> Widget turn on Enable Server Widget switch. Below that, you will find SERVER ID.

Here's how it will look:

screen shot 2018-08-03 at 15 37 33

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,22 @@
 	<br>
 	<p>Desktop App for <a href="https://barterdex.supernet.org">BarterDEX</a></p>
 	<br>
+	<a title="Downloads" href="https://github.com/atomiclabs/hyperdex/releases/latest">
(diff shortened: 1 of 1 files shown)


sindresorhus:
And yes, seems we already have 2K downloads!

lukechilds:
>And yes, seems we already have 2K downloads!

🤯

What is downloads measuring? Release downloads from GitHub? Or git clones? Or some combination of the two?

sindresorhus:
> Release downloads from GitHub?

⬆️

PR #466, opened 2018-08-08
Improve login performance

sindresorhus:
I improved the login performance by running some of the calls concurrently.

I also added detailed timing info to ease future debugging:

```
login
	decrypt: 616.410888671875ms
	swap-db: 6.993896484375ms
	watch-currency-history: 1352.623046875ms
	watch-cmc: 1872.888671875ms
	enable-currencies: 17733.56201171875ms
	watch-currencies: 13172.137939453125ms
	login-total: 32655.123046875ms
```

Notice how `enable-currencies` and `watch-currencies` are very slow. Both of which are calling Marketmaker...

### DIFF
--- a/app/renderer/containers/Login.js
+++ b/app/renderer/containers/Login.js
@@ -12,6 +12,16 @@ import appContainer from './App';
 const config = remote.require('./config');
 const {getPortfolios, decryptSeedPhrase} = remote.require('./portfolio-util');
 
+const setAppWindowBounds = () => {
+	const win = remote.getCurrentWindow();
+	win.setResizable(true);
+	win.setMaximizable(true);
+	win.setFullScreenable(true);
+	win.setMinimumSize(minWindowSize.width, minWindowSize.height);
+	setWindowBounds(config.get('windowState'));
+	win.center(); // TODO: Remove this when `setWindowBounds` handles positioning the window inside the window bounds
+};
+
 const initApi = async seedPhrase => {
 	let url = config.get('marketmakerUrl');
 	if (url) {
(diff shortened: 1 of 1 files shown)


lukechilds:
Brilliant!

This is actually working towards a similar idea I had where the entire login process would be wrapped in an event emitter that would emit textual progress events.

So instead of just showing the static `Logging in...` text for ages we can show the latest text from the event emitter so it would cycle through messages like:

```
Decrypting Portfolio...
Building SwapDB...
Fetching Fiat Prices...
Fetching Historical Prices...
Enabling KMD...
Enabling CHIPS...
Enabling ETH...
Enabling MYTH...
Getting Portfolio Balances...
```

That would give the user a bit more insight as to what's going on rather than maybe them thinking something's crashed if the view doesn't change for a while.

It could also be useful for debugging if we notice enabling a certain currency always takes a while.

Anyway, that's just another idea I had, beyond the scope of this PR, but good to see this refactor is working towards that kind of structure.

lukechilds:
And regarding poor mm performance, it processes each request serially and blocks everything else (as well as getting blocked by anything else that it's currently doing internally).

I believe it also queries Electrum servers pretty inefficiently by opening a new connection for each request as opposed to persisting the connection and getting pushed live updates on balances.

Hopefully, we can push to get this resolved in mmv2.

// cc @ArtemGr @artemii235

sindresorhus:
> This is actually working towards a similar idea I had where the entire login process would be wrapped in an event emitter that would emit textual progress events.

I don't think this is something we should expose to the user. I would rather use the time teaching them how HyperDEX works, by showing an intro the first time, and later some random tips.

PR #467, opened 2018-08-09
Switch to coingecko.com API

### LINKED ISSUE #464: Move from CoinMarketCap
> The Public API will be taken offline on December 4th, 2018. Please migrate your application to the free tier of the Professional API. - https://coinmarketcap.com/api/

They first forced us to move to a completely different v2 API and then months later they deprecate the v2 API in favor of a different paid API. They also don't respond to support requests. Such an amateur show.

---

Relevant Marketmaker issue: https://github.com/artemii235/SuperNET/issues/116

We should investigate https://www.coingecko.com/api/docs/v3

We should also consider setting up our own caching proxy so we don't overload their servers.

sindresorhus:
Fixes #464

It should be faster now too as we can fetch all the data in one request instead of one request per symbol.


app/renderer/containers/App.js
```diff
@@ -21,43 +21,36 @@ const excludedTestCurrencies = new Set([
...
-	if (json.metadata.error) {
-		return fallback;
-	}
+	// The API just ignores IDs it doesn't support, so we need to iterate
+	// our list of symbols instead of the result so we can add the fallbacks.
+	const currencies = symbols.map(symbol => {
+		const currency = data.find(currency => currency.symbol.toUpperCase() === symbol);
+		return {
+			symbol,
+			price: currency ? currency.high_24h : 0,
```
sindresorhus: @lukechilds Is it correct to get `high_24h` or should it be `low_24h`?

lukechilds: I think `current_price` is what we want.

app/renderer/containers/App.js
```diff
@@ -135,9 +128,9 @@ class AppContainer extends Container {
 		}));
 	}
 
-	async watchCMC() {
-		await fireEvery({minutes: 5}, async () => {
-			this.coinPrices = await Promise.all(getCurrencySymbols().map(getTickerData));
+	async watchCoinGecko() {
```
lukechilds: I think maybe we should change this to something more generic like `watchFiatPrice`.

app/renderer/containers/App.js
sindresorhus: Ah yeah, of course. I totally missed that field...

app/renderer/containers/App.js
sindresorhus: 👍

sindresorhus:
@lukechilds Fixed your feedback.

lukechilds:
Perfect.

Do you think it would make sense to use CoinGecko for the historical charts too?

Obvs not urgent at the moment, but would be nice to consolidate everything under one provider, especially if they've reached out to us personally and assured us they have generous limits.

lukechilds:
Good points, I was probably underestimating the effort involved in switching the charts over to a new API.

Issue #181, opened 2018-05-07
Sort and filter swaps in the Trades view


sindresorhus:
Related issue: https://github.com/lukechilds/hyperdex-bugtracker/issues/12

Active trades will end up with a lot of swaps there and it could be useful to sort each column and to be able to filter.

PR #468, opened 2018-08-09
Add ability to sort swaps in Trades view

kevva:
Part of fixing #181. This should also improve performance when rendering large amounts of swaps since only the ones visible will get rendered.

screen shot 2018-08-09 at 18 35 29


sindresorhus:
I think the headers need some kind of `:hover` state to make it clear they can be clicked. And give the headers `{ cursor: default; user-select: none; }`. We don't want the hand cursor in a native app, except for actual links.

sindresorhus:
All the headers are aligned with their column except for the Status header. I think it should be aligned too.

sindresorhus:
The default sort is by date (when you open the Trades view), but it's not indicated in the `Date` header.

app/renderer/components/SwapList.js
```diff
@@ -45,8 +52,39 @@ class CancelButton extends React.Component {
...
+		</div>
+	);
+};
+
+const SwapHeader = props => (
+	<div className="row header">
+		<SwapHeaderColumn {...props} className="timestamp" sortKeys={['timeStarted']}>
+			{t('list.date')}
+		</SwapHeaderColumn>
+		<SwapHeaderColumn {...props} className="pairs" sortKeys={['baseCurrency', 'quoteCurrency']}>
```
sindresorhus: Should this one also sort by date after `baseCurrency` and `quoteCurrency`? So `['baseCurrency', 'quoteCurrency', 'timeStarted']`

app/renderer/components/SwapList.js
```diff
@@ -45,8 +52,39 @@ class CancelButton extends React.Component {
...
+			{t('list.pair')}
+		</SwapHeaderColumn>
+		<SwapHeaderColumn {...props} className="base-amount" sortKeys={['baseCurrency', 'baseCurrencyAmount']}>
+			{t('list.baseAmount')}
+		</SwapHeaderColumn>
+		<SwapHeaderColumn {...props} className="quote-amount" sortKeys={['quoteCurrency', 'quoteCurrencyAmount']}>
+			{t('list.quoteAmount')}
+		</SwapHeaderColumn>
+		<SwapHeaderColumn {...props} className="status" sortKeys={['status']}>
```
sindresorhus: What should `status` sort by for identical statuses?

lukechilds: `timeStarted` I think.

kevva: Would make sense to just sort everyone with `timeStarted` after the others.

sindresorhus:
> The Status column has centered content which makes it hard to align. Should we change it to be left aligned instead?

👍 

> Yeah, technically it is, but by the DB. Felt weird making the component aware of that so I added a prop with a default value instead.

👍 


> Good idea. Used --active-text-color. Dunno if we should use something more subtle?

I think it's a little bit too bright. I would go with this color: `hsla(216, 27%, 80%, 1)`. Also add a slight transition to it, like `transition: color 0.3s`.

I would also add some padding on the text instead of margin, so it's highlighted even when I put my mouse here: 
screen shot 2018-08-14 at 12 55 43

kevva:
> I think it's a little bit too bright. I would go with this color: hsla(216, 27%, 80%, 1). Also add a slight transition to it, like transition: color 0.3s.
> 
> I would also add some padding on the text instead of margin, so it's highlighted even when I put my mouse here:

Fixed. I used `hsla(216, 27%, 20%, 1)` for the Light theme.

app/renderer/components/SwapList.js
```diff
@@ -12,6 +14,11 @@ import './SwapList.scss';
 
 const t = translate(['swap', 'exchange']);
 
+const SortDirections = {
+	ASC: Symbol('asc'),
+	DESC: Symbol('desc'),
```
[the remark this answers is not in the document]
sindresorhus: @lukechilds It's used to implement a pseudo-enum in JS. Slightly better type-safety than strings.

app/renderer/components/SwapList.js
```diff
@@ -69,45 +107,123 @@ const SwapItem = ({swap}) => (
...
+			return <Empty show text={t('list.empty')}/>;
+		}
+
+		const shouldLimit = limit && limit < swaps.length;
+		if (shouldLimit) {
+			swaps = swaps.slice(0, limit);
+		}
+
+		if (sortBy) {
+			swaps = _.sortBy(swaps, sortBy);
```
lukechilds: This is just an alphanumeric sort over the values, right?

If so, that doesn't really make sense for Base/Quote amount because of the different units. The should probably be ordered by their dollar value amount.

lukechilds: Oh, just realised you grouped by base currency unit first, then by amount.

Yeah, that makes sense actually.

app/renderer/components/SwapList.js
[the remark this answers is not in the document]
lukechilds: Interesting, do you know of a good resource where I can read more on that?

lukechilds:
This looks great @kevva, good stuff 🎉 

Interested to hear your thoughts on https://github.com/atomiclabs/hyperdex/pull/468#discussion_r209997229 but happy to merge as is.

Btw, after playing around with the sorting, I'm not sure sorting in asc/desc is that useful for all of the properties, for example, base/quote currency and status.

It would be great if people could set a filter that will limit the selection to things like:

- Only trades with a status of complete|failed|unmatched
- Only trades involving the currency KMD
- Only trades on the trading pair CHIPS/KMD

Obviously out of the scope of this PR, just putting it out there.

app/renderer/components/SwapList.js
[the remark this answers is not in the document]
sindresorhus: http://2ality.com/2016/01/enumify.html

kevva:
> Btw, after playing around with the sorting, I'm not sure sorting in asc/desc is that useful for all of the properties, for example, base/quote currency and status.

Probably not, but I think it's better to stay consistent through all properties. Also, I think that's what most users are familiar with.

> It would be great if people could set a filter that will limit the selection to things like:

Yup, adding filters is next on my list.

app/renderer/components/SwapList.js
kevva: > While that does make sense, I still think maybe total value (fiat equivalent) might be what people are expecting.

Yeah, should probably sort by value first. But would you really expect it to be sorted after the fiat equivalent when it isn't visible anywhere? Maybe you're right though...

sindresorhus: I'm not sure I would expect it to sort by fiat value either. Maybe there should be a separate column for the fiat value of the trade?

lukechilds:
Sorry, when I said sort by amount I meant by fiat amount, not by unit amount.

E.g:

```
1 BTC
500 MKD
5 ETH
```

But I suppose if we do that we should probably also include the fiat amount to make it clear.

E.g:

```
1 BTC ≈ $6,068.83
500 KMD ≈ $490.91
1.5 ETH ≈ $387.885
```

But then that's going to break the layout so maybe just forget about that (sorry)

kevva:
> Sorry, when I said sort by amount I meant by fiat amount, not by unit amount.

Yeah, I know, left a comment in https://github.com/atomiclabs/hyperdex/pull/468#discussion_r210021086. But in case don't use the fiat amount, we should at least use the coin amount instead of the symbol.

sindresorhus:
@kevva But sorting by amounts before symbol is not very helpful. Then you'll have `500 KMD`, `1.5 ETH`, `1 BTC`.

lukechilds:
Yep, I think you should revert to how it was previously, sort by currency, then by amount.

As Sindre suggests a sortable separate column for the fiat value of the trade would be useful and avoid all confusion but not sure if we have enough space to permit that.

sindresorhus:
> As Sindre suggests a sortable separate column for the fiat value of the trade would be useful and avoid all confusion but not sure if we have enough space to permit that.

We do in the Trades view, but not in the compact one in Exchange.

lukechilds:
>We do in the Trades view, but not in the compact one in Exchange.

Perfect.

I'm thinking once GTC orders are implemented we should even change the "Recent Swaps" list to "Open Orders". So it'll only list the in-progress swaps. That will encourage users to use the actual trade history for viewing historical trades. At the moment there is a bit of usage overlap.

This is also how GDAX/Coinbase Pro do it.

sindresorhus:
I think adding a new fiat column should be done in a separate PR though. I would like to ship this for the public launch.

lukechilds:
Yep agree 👍

lukechilds:
@kevva can you revert to sorting by base/quote currency unit, then amount. (sorry)

Then I think this is good to merge.

Good stuff ✨

sindresorhus:
Really nice work, @kevva. This is exactly what I had in mind.

kevva:
🎉

lukechilds:
Yep, spot on @kevva ⭐️

PR #472, opened 2018-08-13
Add donate button

sindresorhus:
It's WIP as I still need to get the KMD and BTC address from Pondsea.

screen shot 2018-08-14 at 00 33 00
screen shot 2018-08-14 at 00 32 56

### DIFF
--- a/app/locales/en-US/app.json
+++ b/app/locales/en-US/app.json
@@ -4,5 +4,9 @@
 		"exampleUrl": "Example: {{url}}",
 		"invalidUrl": "Invalid URL",
 		"title": "App settings"
+	},
+	"donate": {
+		"title": "Donate",
+		"bodyText": "Support the continued development of HyperDEX."
 	}
 }
--- a/app/locales/en-US/app.json
+++ b/app/locales/en-US/app.json
@@ -7,6 +7,6 @@
 	},
 	"donate": {
 		"title": "Donate",
-		"bodyText": "Support the continued development of HyperDEX."
+		"bodyText": "HyperDEX is a free and open source project. You can help support its development by donating to the below addresses."
 	}
 }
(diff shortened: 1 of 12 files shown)


lukechilds:
Looks great.

Could maybe do with a little more text, I'm not sure exactly what to put there but maybe something like:

>HyperDEX is a free Open Source project. You can help support to support HyperDEX development by donating to the following addresses.

lukechilds:
Maybe throw a ❤️ emoji in there somehwere?

sindresorhus:
How about this?

screen shot 2018-08-15 at 00 39 58


Btw, open source should not be capitalized when used in a sentence: https://english.stackexchange.com/questions/157178/capitals-or-not-for-open-source

lukechilds:
That looks great.

>Btw, open source should not be capitalized when used in a sentence

Cool, did not know that, always thought it was like a title of a type of software.

lukechilds:
Maybe move the ❤️ to the beginning of the text to match the first start modal title:

>✨ Thanks for Helping Us Test HyperDEX!

sindresorhus:
> ✨ Thanks for Helping Us Test HyperDEX!

How about we change this one instead to have the emoji at the end? It's more common for emojis to be after the text. They're not icons. And with the heart it looks unaligned when it's in front:

screen shot 2018-08-15 at 01 06 47

lukechilds:
Sure, makes sense.

sindresorhus:
@lukechilds Ok. I'll do that later after this PR. The emoji in the intro modal is in the localized text, so changing the position of it will invalidate all the localizations for the string... I will look for a workaround.

sindresorhus:
Is this good to go now?

lukechilds (approved):

lukechilds:
Beautiful

elevatedmindset209702 (commented):
A lil shaky ground at first but it becomes a little more clearer as you trug along. Understand it's code but what about a glossary or index of some kind for references along the way. Would make some things just a lil 
more ez don't you think?

PR #474, opened 2018-08-15
Add "Add Currency" button in Dashboard for new users

### LINKED ISSUE #425: "Add Currencies" button in the Dashboard
Would be useful to show this if the user has not yet added/remove any currencies, so to encourage them to add some more. Clicking the button would take them to the Settings view.

We can detect that the list of currencies is the same as the default. But what happens the user adds some coins and then removes them so it's back to the default. We wouldn't want to show the button again then. So maybe we should persist if the list of currencies has been edited once? // @atomiclabs/core

kevva:
Feedback on the design welcome :).

button

Fixes #425

### DIFF
--- a/app/locales/en-US/dashboard.json
+++ b/app/locales/en-US/dashboard.json
@@ -30,6 +30,8 @@
 		"value": "{{symbol}} Holdings Value"
 	},
 	"list": {
+		"addCurrency": "Add Currency",
+		"addCurrencyDescription": "You can choose between 100+ currencies!",
 		"emptyWallet": "Empty wallet",
 		"search": "Search…"
 	},
--- a/app/renderer/components/Select.js
+++ b/app/renderer/components/Select.js
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import ReactSelect from 'react-select';
 import './Select.scss';
 
-const Select = ({searchable, clearable, ...props}) => {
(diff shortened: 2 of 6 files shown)


sindresorhus:
I don't think it's clear enough it's a button. I have been meaning to improve the wallet buttons too.

What do you think about this?

screen shot 2018-08-15 at 15 55 49

CSS changes to the button:

```css
background-color: var(--primary-color-opacity-0-2);
border: 1px solid rgba(255, 255, 255, 0.04);
box-shadow: 0 1px 5px rgba(0, 0, 0, 0.3);
border-radius: 8px;
```

lukechilds:
I think I actually prefer the design on @kevva's but I do agree it's not that clear it's a button.

@sindresorhus's looks much more clickable like a button.

sindresorhus:
@kevva Can you experiment with something in between? Maybe what I have, but drop the blue color.

kevva:
I think the former looks better too. Will try yours but with a difference color. It stands out a bit too much I think.

kevva:
Kept the same background and border but added some `box-shadow` and increased the `border-radius`. Is it still not clear that it is a button?

Without hover:

screen shot 2018-08-16 at 18 22 53

With hover:

screen shot 2018-08-16 at 18 23 04

lukechilds:
I think that looks good 👍

sindresorhus:
👍 Let's go with that.

kevva:
Alright, done!

PR #494, opened 2018-08-30
Decouple the swap modal from the swap list

sindresorhus:
And instead of accepting a swap object, it now accepts a swap ID.

The swap modal is now controlled by the swap list instead of controlling itself.

This fixes the issue with an open swap modal suddenly disappearing when the swap completes, as mention in https://github.com/atomiclabs/hyperdex/pull/484.

Notice how the swap list is empty, but the modal is still open:

screen shot 2018-08-30 at 23 55 52

### DIFF
--- a/app/renderer/components/SwapDetails.js
+++ b/app/renderer/components/SwapDetails.js
@@ -1,8 +1,10 @@
 import {remote, clipboard} from 'electron';
 import title from 'title';
 import React from 'react';
+import {Subscribe} from 'unstated';
 import PropTypes from 'prop-types';
 import formatDate from 'date-fns/format';
+import exchangeContainer from 'containers/Exchange';
 import Modal from 'components/Modal';
 import Progress from 'components/Progress';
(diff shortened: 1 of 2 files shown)


app/renderer/components/SwapDetails.js
```diff
@@ -43,11 +45,12 @@ const getOverview = swap => {
 
 class SwapDetails extends React.Component {
 	static propTypes = {
-		swap: PropTypes.object,
+		swapId: PropTypes.string,
+		open: PropTypes.bool,
```
kevva: Nitpick, but I think boolean props should be prefixed with `is` or `has` for clarity.

sindresorhus: I agree, but it's coming from https://github.com/atomiclabs/hyperdex/blob/29ca860171ef90e780d6fbd5317a82a02448266a/app/renderer/components/Modal.js#L21 So the change should be done separately and for all uses.

PR #501, opened 2018-09-04
Use correct value to fix percentage change

lukechilds:
This was broken when we move to CoinGecko.

Will do a followup PR to remove all references to CMC and instead refer to prices as `fiatPrice` instead of `cmcPrice` etc.


sindresorhus:
👍 Looks good.

PR #502, opened 2018-09-04
Add PGP badge to readme linking to Keybase proofs

lukechilds:
Thought this looked cool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 		<img src="https://img.shields.io/github/license/atomiclabs/hyperdex.svg">
 	</a>
 	<a title="Keybase" target="_blank" href="https://keybase.io/hyperdex">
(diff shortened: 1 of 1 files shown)


sindresorhus:
Good idea!

Why is this one much longer than the ones suggested in https://github.com/keybase/keybase-issues/issues/849#issuecomment-198513202 and https://github.com/badges/shields/issues/1299 ?

sindresorhus:
And vote this up: https://github.com/amio/badgen-service/issues/140

lukechilds:
>Why is this one much longer?

32-bit PGP fingerprints are [trivial to brute force](https://evil32.com/) so I used the 64-bit PGP fingerprint.

lukechilds:
Note to self: Update to https://badgen.net, when this is merged: https://github.com/amio/badgen-service/pull/141

PR #516, opened 2018-09-12
Alpha => Beta in translation files

sindresorhus:
Enjoy!


lukechilds:
I've done my best to translate Alpha => Beta in all languages with the help of Google Translate. Might not be perfect in all languages but it's only a small change so should be understandable.

PR #517, opened 2018-09-12
Fetch swap history in `Dashboard`

kevva:
When logging in and viewing the recent activity on the dashboard, it'll show no activity since the trades aren't loaded yet. You have to go to the exchange or trades view for them to show up.

Ideally we would run the following code as soon as we're logged in https://github.com/atomiclabs/hyperdex/blob/master/app/renderer/containers/Exchange.js#L33-L46.

### DIFF
--- a/app/renderer/containers/Dashboard.js
+++ b/app/renderer/containers/Dashboard.js
@@ -1,7 +1,8 @@
 import _ from 'lodash';
 import pMap from 'p-map';
-import {Container} from 'unstated';
 import appContainer from 'containers/App';
+import exchangeContainer from 'containers/Exchange';
+import SuperContainer from 'containers/SuperContainer';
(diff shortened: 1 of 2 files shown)


sindresorhus:
I'm aware of the problem. I've been meaning to refactor the swap history to be in AppContainer instead, but wanted to wait for the refactoring work @lukechilds commented he was working on a while ago, so not to conflict.

lukechilds:
Which work was that, the cmc/fiat thing or something else?

sindresorhus:
Yes, the fiat thing.

lukechilds:
Ah, sorry, that got `git stash`ed in the mad rush for beta. I didn't think it was important for beta and didn't want to make a load of code changes just before the release.

I didn't make much progress on it.

PR #531, opened 2018-09-24
Fix `Document update conflict` error

### LINKED ISSUE #527: Unhandled Promise Rejection
Unfortunately I don't know how related these two issues are to one another but I placed two sell orders, one with KMD for HUSH and another with KMD for GUSD. Neither of the two orders are appearing in order book or the charts, but they are appearing as open orders with order ID's (I'll include at bottom) in the "Open orders" section on "exchange" tab as well as "Trades" tab. 

I left these orders as they are, left the app open (not minimized), and was doing other things when I noticed the HyperDEX icon flashing orange. I opened it I had the following error:
image

Copy and pasted it into Discord, and made no changes to my trades or changed tabs etc. left the app open (again not minimized) and continued working on other things, and saw it flashed orange again with the same error. Having closed out the error (by pressing copy error) the error immediately reappears, I then closed out by pressing OK, and it immediately reappeared again. I closed out a third time (with OK) and it hasn't appeared yet (as I write this).

I'm not sure how this error came about but it would seem related to the two orders I placed, and perhaps related to why they are not showing up in the chart or order book. Here are the order IDs:

KMD/HUSH sell order: be0cd0ebafc3d9bba4b1f8372e2c4f0336f0a0804c3d5e063ba958f3d06c007e 
KMD/GUSD sell order: 12ade51d1d634c4aee497681d77bd764e43f96a9bea9dde5e88dc56da822469f

Note: I attempted to replicate this error with zero open orders but so far haven't received any error message.

---

HyperDEX 0.1.3
Electron 3.0.0
win32 10.0.17134
Locale: en-US

### LINKED ISSUE #528: Swap error
After creating an order I receive the below errors repeatedly.

Version: 0.1.3
OS: win 10
Errors: below
Severity: High

image

image

sindresorhus:
Fixes #527
Fixes #528

We were not correctly updating the PouchDB documents. There is a required ceremony of updating `_rev` that we were not doing. I think the issue only first showed up now as IndexedDB either got faster or something changed internally with the Electron 3 upgrade.

See: https://pouchdb.com/guides/updating-deleting.html

I'm using `pouchdb-upsert` which handles some of the complications of updating a document.

### DIFF
--- a/app/renderer/swap-db.js
+++ b/app/renderer/swap-db.js
@@ -1,5 +1,6 @@
 import PouchDB from 'pouchdb-browser';
 import pouchDBFind from 'pouchdb-find';
+import pouchDBUpsert from 'pouchdb-upsert';
 import cryptoPouch from 'crypto-pouch';
 import Emittery from 'emittery';
 import PQueue from 'p-queue';
@@ -13,6 +14,7 @@ import {translate} from './translate';
 const t = translate('swap');
 
 PouchDB.plugin(pouchDBFind);
+PouchDB.plugin(pouchDBUpsert);
 PouchDB.plugin(cryptoPouch);
 
 class SwapDB {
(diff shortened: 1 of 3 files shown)


sindresorhus:
I tested the beta and got the `Conflict update error` too. With these changes, I built for production, ran it, and now got a swap through without any errors, so seems to be working.

lukechilds:
Awesome good catch!

Sorry, totally missed that, assumed `db.put()` was completely safe for updating.

Issue #302, opened 2018-05-23
Can not withdraw komodo coin.


VONLUEAPS:
![image](https://user-images.githubusercontent.com/16932261/40427397-8f55ffae-5ed8-11e8-8fd5-15657a625437.png)

Unhandled Promise Rejection
Error: Couldn't create withdrawal transaction:
{ txid: '0000000000000000000000000000000000000000000000000000000000000000',
  txfee: 1000,
  complete: false }
    at errorWithObject (app://-/bundle.js:90100:46)
    at app://-/bundle.js:90357:11
    at Generator.next (<anonymous>)
    at step (app://-/bundle.js:90089:191)
    at app://-/bundle.js:90089:361
    at <anonymous>

This is because the transmission fee (0.00001 KMD) was not calculated when you clicked the Max button.

lukechilds:
Thanks,yeah we're aware of this, we need to properly handle this.

Only issue is it works completely differently with Bitcoin because it uses dynamic fees. It would be helpful if the API response gave us a reason why creating the TX failed, and if it's value related, what the max values we can use are.

@jl777 or @artemii235, is this something that would be simple to add?

sindresorhus:
@artemii235 :arrow_up:

artemii235:
Please check mm console log when you get errors, it might give better description. It worth to add detailed error description to every API call, however it's not really possible in MM 1.0. MM2.0 will provide better error messages.

sindresorhus:
@artemii235 Looking at the mm logs might help us understand it, but it wouldn't help us differentiate between errors. We need better API responses.

n41r0j:
This let's you set the txfee manually, even for BTC. When you want to withdraw the complete balance, make the amount as defined in the outputs array subtract the txfee. 

TX fee is in satoshis.

```text
{
    "method": "withdraw",
    "coin": "KMD",
    "txfee":5000,
    "outputs": [
        {
            "Raddress": <MAX> - txfee
        }
    ]
}```

lukechilds:
@n41r0j but we have no idea what fee to set because we don't know how big the transaction is. If we set a fixed amount we could easily end up with a TX that massively over paid or that hours to confirm.

n41r0j:
[...]

PR #533, opened 2018-09-25
Warn about withdrawing the whole balance

sindresorhus:
Since it doesn't seem like #302 will be fixed anytime soon and lots of users are encountering it, we should make the issue clear and prevent it from happening.

screen shot 2018-09-25 at 15 44 55

@lukechilds Is the text correct?

### DIFF
--- a/app/renderer/views/Dashboard/WithdrawModal.js
+++ b/app/renderer/views/Dashboard/WithdrawModal.js
@@ -3,7 +3,6 @@ import roundTo from 'round-to';
 import Modal from 'components/Modal';
 import Button from 'components/Button';
 import Input from 'components/Input';
-import Link from 'components/Link';
 import appContainer from 'containers/App';
 import dashboardContainer from 'containers/Dashboard';
 import {formatCurrency} from '../../util';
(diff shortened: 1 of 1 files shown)


lukechilds:
The issue is a bit more complex than that. But without going to far down the rabbit hole, yeah that's the general idea.

We can actually resolve this with most alts and ACs as they have a hardcoded txfee so we can just reduce this from the amount. However I was pretty stubborn about not implementing this as it would mean we would need completely different withdrawal logic for different currencies.

I'd much rather we find the time to put the work in to get proper withdrawals working with fees and use the same code for all currencies.

Off the top of my head, BTC, ETH and all ERC20 tokens use dynamic fees so will require a fair bit of work to integrate properly (probably not possible in a user friendly way with current mm). However, all other currencies would work fine using their hardcoded txfee and subtracting that from the available amount.

Also, I'm not convinced this is the only cause of the error we get from withdrawals. We seem to get lots of errors in other scenarios too, unrelated to large balances but the error response gives us no clue as to what this is, they are all the same.

**TL;DR Yeah, I'd say we should merge this to hopefully reduce the confusion and amount of support requests we get regarding it, but it's not really explaining the full picture.**

PR #534, opened 2018-09-25
Persist the logged-in window state

sindresorhus:
If the user quits the app or logs out, the next time they log in, the window is restored to the same position.

### DIFF
--- a/app/renderer/containers/Login.js
+++ b/app/renderer/containers/Login.js
@@ -18,8 +18,13 @@ const setAppWindowBounds = () => {
 	win.setMaximizable(true);
 	win.setFullScreenable(true);
 	win.setMinimumSize(minWindowSize.width, minWindowSize.height);
-	setWindowBounds(config.get('windowState'));
-	win.center(); // TODO: Remove this when `setWindowBounds` handles positioning the window inside the window bounds
+
+	const windowState = config.get('windowState');
+	setWindowBounds(windowState);
+	const hasPositionState = Reflect.has(windowState, 'x');
+	if (!hasPositionState) {
+		win.center();
+	}
 };
 
 const initApi = async seedPhrase => {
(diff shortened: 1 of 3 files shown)


app/renderer/containers/Login.js
```diff
@@ -18,8 +18,13 @@ const setAppWindowBounds = () => {
...
 	win.setFullScreenable(true);
 	win.setMinimumSize(minWindowSize.width, minWindowSize.height);
-	setWindowBounds(config.get('windowState'));
-	win.center(); // TODO: Remove this when `setWindowBounds` handles positioning the window inside the window bounds
+
+	const windowState = config.get('windowState');
+	setWindowBounds(windowState);
+	const hasPositionState = Reflect.has(windowState, 'x');
+	if (!hasPositionState) {
+		win.center();
```
lukechilds: Why do you set both setWindowBounds(windowState); and win.center(); wouldn't it be more efficient to just do one or the other?

Or does setWindowBounds(windowState); do some other stuff that we need in both boolean branches so it's just easier to call that first and then overwrite x/y coords.

lukechilds:
Also, not directly related to this PR but slightly relevant.

I think we should increase the auto resize dimensions after login. The UI looks much better when the swap list is full width and most users should probably use the app maximised.

If you think about centralised exchanges, most users would have it open in a maximised browser tab. We're trying to fit a similar amount of information in quite a small amount of screen space.

app/renderer/containers/Login.js
sindresorhus: The first time the app launches, the `windowState` only has width/height, not x/y. That's what `hasPositionState` checks. So if there's no position set, we center the window as it's nice for it to be at the center for the first launch. So the first launch it does two calls yes, but they're very cheap and `setWindowBounds` only sets the dimensions then, while `win.center()` set's the position.

PR #544, opened 2018-10-08
Add filters to trade history

kevva:
As the title says, this adds filters for date, symbol pairs and order type. Suggestions on how the design could be improved are welcome.

Also, the date input currently throws if you enter anything due to an [internal function](https://github.com/gpbl/react-day-picker/blob/master/src/DayPickerInput.js#L393-L419) in `react-day-picker` which expects an event as the first argument `onChange`, but we give it `(value, event)`. Not sure how we like to handle that.

screen shot 2018-10-08 at 02 36 02


sindresorhus:
Looks very good to me. Just some minor nitpick:

---

If I write in the "Pair" input, the text is black and with too much left padding:

screen shot 2018-10-08 at 16 07 40


---

When hovering over the `<` and `>` in the calendar, it should not use a hand pointer.

---

The `v` here looks one pixel unaligned vertically:

screen shot 2018-10-08 at 16 12 50

---

If I select a date in "From" and "To" calendar, then open the "To" calendar and select a day inside the range, it doesn't update "To", but rather clears it.

---

Select `2018-10-18` in "From" and open the "To" calendar. Then try hovering over today's date and it disappears. It should have no hover effect like the rest of the disallowed dates.

webpack.config.js
```diff
@@ -44,8 +44,8 @@ module.exports = (env, options) => ({
 				options: require('./package.json').babel,
 			},
 			{
-				test: /\.scss$/,
-				exclude: /node_modules/,
+				test: /\.(s)?css$/,
```
sindresorhus: I would prefer `test: /\.(css|scss)$/,` for readability.

kevva:
> When hovering over the < and > in the calendar, it should not use a hand pointer.

Fixed. Removed the pointer cursor from the calendar days too.

> The v here looks one pixel unaligned vertically:

Indeed. I moved the arrow up one pixel.

> If I select a date in "From" and "To" calendar, then open the "To" calendar and select a day inside the range, it doesn't update "To", but rather clears it.

I wanted it to clear upon clicking on the current date, not on the dates in range. Seems like `clickUnselectsDay` take all days in the range into account. Removed it for now.

> Select 2018-10-18 in "From" and open the "To" calendar. Then try hovering over today's date and it disappears. It should have no hover effect like the rest of the disallowed dates.

Fixed.

sindresorhus:
The date format should be `2018-10-02`, not `2018-10-2`

sindresorhus:
Can you upgrade to latest Yarn version and run `$ yarn`? I get some missing integrity hashes in the lockfile when I checkout this PR.

kevva:
> The date format should be 2018-10-02, not 2018-10-2

For all locales?

sindresorhus:
I think the square that appears when you hover over days in the calendar should have `2px` in border-radius.

sindresorhus:
- [ ] Validation for the manual input in the date fields.

sindresorhus:
Thinking about this some more, I think it would be better to add default values there instead of placeholders. We could show `To` as today, and `From` as up to one year ago, or the date of the oldest entry if it's less than one year. That way it's clear to the user the format they need to use. We should also validate the input and only allow the pattern `/\d{4}-\d{2}-\d{2}/`.

sindresorhus:
I also don't think you should be able to select a newer date than today in the calendar, by disabling newer dates.

kevva:
> I also don't think you should be able to select a newer date than today in the calendar, by disabling newer dates.

Agree. Also, is there a point to keep older dates than the oldest swap or should we disable those too?

kevva:
> It needs to handle the case where it's close to one of the ends.

To my knowledge, we can't target the element immediately before a selector like we can with elements immediately after (using `+`). Suggestions on how we solve that?

kevva:
That would select all elements before it, not just the one next to it I think.

sindresorhus:
Can you fix it in JS by adding a special class then they're together. Either that or drop the border radius, but I think it looks better with a slight border radius, so would prefer a solution than just dropping it.

sindresorhus:
@kevva Can you fix the merge conflict?

sindresorhus:
@kevva The `Date:` positioning is off because of my changes in https://github.com/atomiclabs/hyperdex/commit/5768b3ef897649b6d1853e5634b17f3b5ed5993a.

screen shot 2018-10-29 at 14 51 25

sindresorhus:
@kevva I forgot, we also need to validate that the input is an actual valid date, not just the pattern. `2018-10-50` is a valid pattern, but not date. We could either make `pattern` support a function too or add a `type=date`. The check could be something like `moment('2018-10-50').isValid()`.

sindresorhus:
Currently, it's possible to to manually write `2019-10-24` in the `From` field. I don't think that should be allowed. How about we make it an error (red input) for a second and then revert it to to today's date?

Is it possible to prevent the `From` visual calendar from showing future dates?

kevva:
> I forgot, we also need to validate that the input is an actual valid date, not just the pattern. 2018-10-50 is a valid pattern, but not date. We could either make pattern support a function too or add a type=date. The check could be something like moment('2018-10-50').isValid().

Hmm, I think accepting a function is the cleanest and most scalable way of doing it here. Will fix.

> Currently, it's possible to to manually write 2019-10-24 in the From field. I don't think that should be allowed. How about we make it an error (red input) for a second and then revert it to to today's date?

Shouldn't it follow the same behaviour as the `To` input? I.e. that it gets a shake animation and then reverts the date.

> Is it possible to prevent the From visual calendar from showing future dates?

Yes.

sindresorhus:
> Shouldn't it follow the same behaviour as the To input? I.e. that it gets a shake animation and then reverts the date.

Yeah, that's better. Just make sure it only reverts when the pattern is complete. You wouldn't want this behavior: User writes `2018-02`, then stops to look up the day, then goes back and it's all gone.

>> Is it possible to prevent the From visual calendar from showing future dates?

> Yes.

Let's do it then.

kevva:
Should we do it for the `To` calendar too? No point in showing future dates if they're all disabled.

sindresorhus:
> Should we do it for the To calendar too?

👍

sindresorhus:
Tiny nitpick, but can make the endpoint here slightly less round? I looks slightly uneven now because of the large border radius.

screen shot 2018-10-31 at 17 28 12

sindresorhus:
@kevva If I enter `2018-10-40` in `From`, and then blur the input, it still shows `2018-10-40`. When I focus it again, it changes back to `2018-10-30`. That's some weird behavior.

sindresorhus:
If I enter `2018-10-99999999` in `From`, then blur the input, the input becomes red. I then select a valid date, and the input continues to be red.

kevva:
> If I enter 2018-10-40 in From, and then blur the input, the time it takes to reset is too long. It feels almost like a mistake when it suddenly changes a second later. It should revert right after the shake animation.

Yeah, agree. I'll lower the timeout slightly.

> If I enter 2018-10-99999999 in From, then blur the input, the input becomes red. I then select a valid date, and the input continues to be red.

@sindresorhus, yup. That's what's left. Problem is that `onChange` isn't triggered in `Input` when entering an invalid date in `react-day-picker`, thus why we have to update the internal state of it [manually](https://github.com/atomiclabs/hyperdex/blob/fc3de311762bffc74b06e8cfaea4910d7ccfb543/app/renderer/components/DateInput.js#L45-L46). Probably because of https://github.com/gpbl/react-day-picker/issues/815. I think I have a workaround for it though.

sindresorhus:
The validation/reset behavior feels good now 👌

sindresorhus:
Really nice work on this, @kevva. It turned out great! And it's going to be very useful for people doing a lot of trades.

PR #546, opened 2018-10-08
Improve "Add a Currency" docs

sindresorhus:
Feedback welcome on the original PR too: https://github.com/atomiclabs/hyperdex/pull/540

I intend to keep this one open for a while to gather feedback.


sindresorhus:
- [ ] Requirement: The submitter should commit to keeping the Electrum servers running and if they need to be taken down, give us due notice so we can update/remove.

PR #555, opened 2018-10-15
Move the swap history data to the AppContainer

sindresorhus:
Since the data is used by multiple views. This simplifies a lot of existing code.

### DIFF
--- a/app/renderer/containers/App.js
+++ b/app/renderer/containers/App.js
@@ -6,7 +6,8 @@ import _ from 'lodash';
 import Cycled from 'cycled';
 import coinlist from 'coinlist';
 import roundTo from 'round-to';
-import {Container} from 'unstated';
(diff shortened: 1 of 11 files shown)


app/renderer/containers/App.js
```diff
@@ -111,6 +152,8 @@ class AppContainer extends Container {
 			activeView: 'Dashboard',
 			portfolio,
 		});
+
+		this.initSwapHistoryListener();
```
kevva: Should this run on every view? Isn't it enough to run it once on mount?

sindresorhus: What do you mean by "every new view"? `logIn()` is only called once when done logging in.

kevva: Yeah, sorry. Didn't look at the line numbers so it looked to my like this was inside `setActiveView`.

PR #557, opened 2018-10-18
Make the swap list rows clickable

sindresorhus:
Instead of using the tiny "View" button. There is now a hover effect to make it clear that it's clickable. This is generally how it works natively and after using it for a few days, I feel it's better.

screen shot 2018-10-18 at 23 33 26

The "Cancel" button is still clickable.


kevva:
I like it!

sindresorhus:
I have fixed the spacing.

PR #563, opened 2018-10-31
Marketmaker v2 test

sindresorhus:


sindresorhus:
I had to leave out the Linux binary as it's a whopping 114 MB, so GitHub rejected it.

@artemii235 Is the size a mistake?

```
remote: error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.        
remote: error: Trace: 46e95591ec776c09236d421ddef22b70        
remote: error: See http://git.io/iEPt8g for more information.        
remote: error: File app/marketmaker/bin/linux/marketmaker is 113.95 MB; this exceeds GitHub's file size limit of 100.00 MB        
To https://github.com/atomiclabs/hyperdex
 ! [remote rejected]     mm-v2-test -> mm-v2-test (pre-receive hook declined)
error: failed to push some refs to 'https://github.com/atomiclabs/hyperdex'
```

sindresorhus:
I managed to get HyperDEX to connect to Marketmaker, but when HyperDEX calls the `getendpoint` command, we get an error back: `{error: "Can't queue the command when ws endpoint is disabled!"}`.

@artemii235 Is WebSocket support implemented in v2?

app/renderer/api.js
```diff
@@ -72,6 +76,13 @@ export default class Api {
 			throw error;
 		}
 
+		console.log('Result:', result);
+
+		// Some requests, like the `portfolio` command, have 200 status code even when it returns an error...
+		if (result.error) {
+			throw new Error(result.error);
+		}
```
sindresorhus: @artemii235 Is it intentional that errors still have status code 200? If not, I'm happy to open an issue about it.

artemii235:
> I had to leave out the Linux binary as it's a whopping 114 MB, so GitHub rejected it.
> 
> @artemii235 Is the size a mistake?
> 
> ```
> remote: error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.        
> remote: error: Trace: 46e95591ec776c09236d421ddef22b70        
> remote: error: See http://git.io/iEPt8g for more information.        
> remote: error: File app/marketmaker/bin/linux/marketmaker is 113.95 MB; this exceeds GitHub's file size limit of 100.00 MB        
> To https://github.com/atomiclabs/hyperdex
>  ! [remote rejected]     mm-v2-test -> mm-v2-test (pre-receive hook declined)
> error: failed to push some refs to 'https://github.com/atomiclabs/hyperdex'
> ```

We use debug builds to get full stack traces in case of panics. So the size doesn't look like mistake. However I'll check if size of Linux nightly build can be reduced.

artemii235:
> I managed to get HyperDEX to connect to Marketmaker, but when HyperDEX calls the `getendpoint` command, we get an error back: `{error: "Can't queue the command when ws endpoint is disabled!"}`.
> 
> @artemii235 Is WebSocket support implemented in v2?

Did you set `queueid` param in `getendpoint` call? Please provide full request body.

app/renderer/api.js
artemii235: We want to avoid such situation in MM2. The RPC methods that are not fully ported to Rust will behave this way for some time. I've created an issue about it: https://github.com/artemii235/SuperNET/issues/220.

ArtemGr:
> I had to leave out the Linux binary as it's a whopping 114 MB, so GitHub rejected it.

We can pack the binaries with something like the upx, in order for them to fit the GitHub commits,

but I wonder whether it's wise to be committing binaries into a version control system in the first place?
Is it the only way to do a MM2[-nightly] HyperDex version?

sindresorhus:
> Did you set queueid param in getendpoint call? Please provide full request body.

Yup, the call is exactly the same as for mm v1 where it works.

Request body:

```js
{queueid: 0, needjson: 1, method: "getendpoint", port: 62636, userpass: "<redacted>"}
```

Response:
```js
{type: "cors", url: "http://127.0.0.1:62422/", redirected: false, status: 200, ok: true, …}
```

Response body:

```js
{error: "Can't queue the command when ws endpoint is disabled!"}
```

sindresorhus:
> but I wonder whether it's wise to be committing binaries into a version control system in the first place?
Is it the only way to do a MM2[-nightly] HyperDex version?

Probably not the best idea, but it has had some benefits. By committing the binaries we could be more sure they are not tampered. For example, if we were to automatically download them at build time, we would have to implement signature checking. Also, the binaries for mm v1 are not that big and we didn't update very often, so it was not a problem then. However, even the mm v2 release binaries are probably going to be bigger than mm v1 because of Rust, so we should indeed find a better solution for this. Suggestions welcome!

ArtemGr:
> `{error: "Can't queue the command when ws endpoint is disabled!"}`

This error was introduced here: https://github.com/artemii235/SuperNET/commit/6550c0eae34ae8da1162b8a49bcf03060b654314#diff-087cef82ec3a4f92264696f790723461R2006

@artemii235 will probably fix it. Meanwhile, I suspect that you don't need the `"queueid": 0` part when enabling the WebSocket endpoint. For example, the following query works:

    curl --url "http://127.0.0.1:7783" --data "{\"userpass\":\"...\",\"method\":\"getendpoint\",\"needjson\": 1,\"port\":123}"

And the `queueid` parameter starts to work when the WebSocket endpoint is thus enabled.

sindresorhus:
@ArtemGr Thanks, that worked around the issue. I'm now able to login. I still could not get the order books to show up though, even though I was connected to peers, but not sure whether that's a mm v2 problem or just low market activity.

ArtemGr:
I've tried UPX and unfortunately it interferes with the crash handling as well. That is, if we strip the binary (to 15 MiB) or compress it with UPX (to 27 MiB) we loose the ability to generate the stack traces on crash.

It's curious how the Windows build is different: there we can remove the ".pdb" file (of 71 MiB) yet still retain the ability to generate the stack traces.

> Probably not the best idea, but it has had some benefits. By committing the binaries we could be more sure they are not tampered. For example, if we were to automatically download them at build time, we would have to implement signature checking.

Where this no-tampering guarantee is coming from?
The way I see it, somebody needs to download and commit the binaries and they can be tampered at that moment the same way they can be tampered at the moment of uploading them into release. In particular they can be tampered: 1) on the server you're getting them from; 2) with a man-in-the-middle attack when you're downloading them; 3) by a virus on your computer; 4) with a man-in-the-middle attack when you're committing them; 5) on GitHub; 6) with a man-in-the-middle attack when users are downloading the files from GitHub; 7) by a virus on the user's computer. Some of these steps are protected by weak TLS encryption and the like, but how would we know that they are better protected than the binaries we upload into GitHub releases?

And how direct upload of binaries from the build server and into GitHub, where we have less angles of attack, is less secure?

Are you somehow assuming that Git history is tamper-free?

> Also, the binaries for mm v1 are not that big and we didn't update very often, so it was not a problem then.

Even if they are small, they add up in the history, so it's actually a problem if you would try to pull the repository with all this accumulated history inside.
cf. https://homes.cs.washington.edu/~mernst/advice/version-control.html#Dont_commit_generated_files

> Suggestions welcome!

We can generate smaller binaries by lowering the `debuginfo` level in the released version.
(We might get even smaller with https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#config-profiles in the future).
Lowering `debuginfo` from 2 to 1 reduces the size of the MM2 binary to 80 MiB, this should fit the [100 MB quota](https://help.github.com/articles/what-is-my-disk-quota/).
I'm going to work on patching the [Linux] CI build.

But the idiomatic way to handle this is to upload binaries into the GitHub releases (step 7 in https://help.github.com/articles/creating-releases/, can be automated with https://developer.github.com/v3/repos/releases/#upload-a-release-asset).

There's also an idea that with the Hyper-MM2-nightly release we maybe want to test the latest MM2 always, so it might make sense to create a small binary loader to be included with HyperDEX that would fetch the freshest MM2 binary from the nightly builds server.

ArtemGr:
So locally this option helps with the Linux binary size (while still allowing the stack traces to work): https://github.com/artemii235/SuperNET/commit/6e16ed5b31e1cf6bacf3df425113f4b3380221ec. Waiting for the CI build to finish.

sindresorhus:
I noticed that mm v2 is significantly slower to login than mm v2.

These two Electrum calls are for ETOMIC:


mm v1:

```
API method: electrum (1.564s)
API method: electrum (0.405s)
```

mm v2

```
API method: electrum (17.398s)
API method: electrum (34.909s)
```

artemii235:
@sindresorhus MM2 doesn't contain latest fixed from MM1 master branch. I guess it's caused by this issue: https://github.com/artemii235/SuperNET/issues/162. I'll repeat the fix in MM2 and also fix the `queueid: 0` issue to keep backwards compatibility.

sindresorhus:
Regarding the empty order books. One thing I've noticed is that we actually get some bids and asks with mm v2, it's just that `numutxos`, `avevolume`, `maxvolume`, `depth`, are always `0`. With mm v1, they have values.

##### mm v1

```json
{
  "bids": [],
  "numbids": 0,
  "biddepth": 0,
  "asks": [
    {
      "coin": "CHIPS",
      "address": "RT9MpMyucqXiX8bZLimXBnrrn2ofmdGNKd",
      "price": 0.02485921,
      "numutxos": 148,
      "avevolume": 0.0596621,
      "maxvolume": 5.65199009,
      "depth": 8.82999156,
      "pubkey": "bab6ad2eebe1e666369cab504d4622b22c1f1ef718ef388e88020f30a1573e01",
      "age": 3,
      "zcredits": 0
    }
  ],
  "numasks": 1,
  "askdepth": 8.82999156,
  "base": "CHIPS",
  "rel": "KMD",
  "timestamp": 1541141060,
  "netid": 0
}
```

##### mm v2

```json
{
  "bids": [
    {
      "coin": "KMD",
      "address": "RB8yufv3YTfdzYnwz5paNnnDynGJG6WsqD",
      "price": 0.01577993,
      "numutxos": 0,
      "avevolume": 0,
      "maxvolume": 0,
      "depth": 0,
      "pubkey": "8adbf9d5de4fb49ec1fca9ca3f28ba384715752de1197c9cddbb756c3d2a1a7c",
      "age": 4,
      "zcredits": 0
    }
  ],
  "numbids": 1,
  "biddepth": 0,
  "asks": [
    {
      "coin": "CHIPS",
      "address": "RB8yufv3YTfdzYnwz5paNnnDynGJG6WsqD",
      "price": 0.02479704,
      "numutxos": 0,
      "avevolume": 0,
      "maxvolume": 0,
      "depth": 0,
      "pubkey": "8adbf9d5de4fb49ec1fca9ca3f28ba384715752de1197c9cddbb756c3d2a1a7c",
      "age": 4,
      "zcredits": 0
    }
  ],
  "numasks": 1,
  "askdepth": 0,
  "base": "CHIPS",
  "rel": "KMD",
  "timestamp": 1541140648,
  "netid": 0
}
```

artemii235:
@sindresorhus I've recently removed UTXO related code from orderbook. The reason is that we need to remove the knowledge about specific coins implementation from other application layers (and from GUI too). It's very hard to add non-UTXO coins when everything is dependent on UTXOs. I can hardcode numutxos and other info to some values but we can mark these fields as "Deprecated".

lukechilds:
I strongly agree with your decision that it makes sense to move the API requests and response away from low lever BTC specific parameters like UTXOs and instead use a higher level of abstraction such as "amounts" etc to allow for better integration with other cryptocurrencies with completely different architectures.

E.g UTXO based accounts (BTC) and balance based accounts (ETH) should be a low-level implementation detail, abstracted away and hidden from the top level API.

However, if this is the case mmv1 -> mmv2 is not going to be a simple swap-in-out job, it may take quite a bit of refactoring.

artemii235:
Ok, here is queueid and electrum fix: https://github.com/artemii235/SuperNET/commit/e793b5b76cdab7d2a6a961d1249896584e54d6e4. Utxos hardcode in order book will take bit more time to do, is it also possible to remove the `utxos > 0` requirement to display the order in GUI?

sindresorhus:
@artemii235 Re UTXO, yes, I already did the change: https://github.com/atomiclabs/hyperdex/pull/563/commits/fcb7b77da41f9c9c80f6a13bffcf6424fa3214bc Order books show up now.

sindresorhus:
> Utxos hardcode in order book will take bit more time to do

I don't think you need to do that change. As long as it's easy for us to workaround we can do that.

sindresorhus:
How about `avevolume`, `maxvolume`, and `depth`?

ArtemGr:
> E.g UTXO based accounts (BTC) and balance based accounts (ETH) should be a low-level implementation detail, abstracted away and hidden from the top level API.

There might be a value in providing the user with the information about her UTXOs though. When I'm using online banking, I can check my current balance with but a glance, but I can _also_ get a detailed list of transactions, credits and pending operations. I think that here we might want to provide similar desalinization options, given that UTXOs will still be used for the trades, will no doubt be mentioned in articles and in-depth guides and reflected in the blockchain exploration tools. The important thing is for the information in the exchange UI (HyperDEX) is to match the information the user has from other sources (wallets, blockchain explorers), in order to increase her confidence in the software and trades. High-level APIs are good, but transparency is also important, it will be great if we could strike the right balance between the two.

artemii235:
Detailed list of transactions can be also provided without leaking the underlying technology implementation. I bet that 95% of crypto users doesn't know what UTXO is. HyperDEX claims to be "Grandma friendly" atomic swaps, so the app should be as simple as possible.

ArtemGr:
> I didn't say no-tampering, only less chance, but you're right that it's still a lot of chances of tampering, so I guess my argument is moot.

Actually, in software security all we have are chances and costs. Any system hypothetically be breached, the question is only the cost. What I'm not convinced of though it that committing binaries makes them somehow harder to tamper with.

For example, let's say hacker H finds a way to tamper with GitHub commits, then publishes her findings somewhere. There will be a time window, the length of which depends on where this information was published, that would allow attacker A to tamper with committed binaries at a low cost.

My point is, we shouldn't assume that something is more secure or more costly to crack simply because it's more complex. Especially so when we're talking about a public service such as GitHub.

> I think the way forward is for you to have your build server push new mm v2 builds and checksums to GitHub Releases and we can add a build script to HyperDEX that fetches releases and validates them at build time. Thoughts?

The way I see it, MM2 is considered experimental software and what we're doing here is making a version of HyperDEX that will make it easier for everyone to test it. That is, we're not making a stable MM2 release _yet_.

And maybe the easiest way to test an experimental software, and a well-established practice at that, is to have the nightly builds of it. We have nightly Rust builds, nightly Firefox builds, nightly Linux builds etc. Using the nightlies shortens the distance between the development and the testing, often making it easier to pinpoint and narrow down the problems discovered by the early testers.

And we don't want the nightly builds in GitHub releases: the nightlies and the release deployment practices are different, we neither want to spam GitHub with nightlies nor do we want to suffer a hit from the simultaneous existence of a thousand of nightly releases. What we want is a version of HyperDEX that just works with the latest nightly.

How about we release an early testers version of HyperDEX that is somewhat tailored to MM2 and contains a script that automatically fetches or updates the latest MM2 binary from http://195.201.0.6/etomic/?

> Our master branch is protected, so it cannot be force-pushed.

I assume that what we're talking about is protecting the binaries from the attacks by malicious third parties. Attacker can, hypothetically speaking, obtain administrative privileges (GitHub has been hacked to this end before), disable the master branch protection, tamper with the Git history, then re-enable the protection. GitHub has an impressive set of defensive options and processes, but sometimes a complex system can be cracked with a simple well-placed attack, not unlike blowing up the Death Star in Star Wars. It is just my opinion, but I don't think there's any proof that GitHub commits are more tamper-proof than GitHub release binaries uploaded otherwise.

ArtemGr:
> I bet that 95% of crypto users doesn't know what UTXO is. HyperDEX claims to be "Grandma friendly" atomic swaps, so the app should be as simple as possible.

Did we test it on real grandmothers already?

artemii235:
No, and I'm sure the test will fail :slightly_smiling_face: It's sometimes hard even for experienced IT persons.

ArtemGr:
There is a point where making the interface simpler actually makes it less obvious to the users. You're probably right in that UTXOs aren't well-known, but an option to "see the balance in the blockchain", linking to the blockchain explorer displaying the UTXOs, might give a user some confidence by connecting some of the dots.

sindresorhus:
> For example, let's say hacker H finds a way to tamper with GitHub commits

Let's not start discussing wild hypotheticals. If someone breaches the GitHub protected branch feature, there are many other projects much worse affected by that than us. What if someone tampers with the Rust compiler, or you know the CPU ;) There's always a risk, and I do agree, committing does not give much extra security benefit.

> And maybe the easiest way to test an experimental software, and a well-established practice at that, is to have the nightly builds of it. We have nightly Rust builds, nightly Firefox builds, nightly Linux builds etc. Using the nightlies shortens the distance between the development and the testing, often making it easier to pinpoint and narrow down the problems discovered by the early testers.

We have nightly builds builds too (https://github.com/atomiclabs/hyperdex-nightlies) and we could make the new build script use the latest mm v2 build available at build-time.

> And we don't want the nightly builds in GitHub releases

Use a separate repo for it like we do? We would prefer a secure and trusted endpoint to fetch mm v2 builds.

> How about we release an early testers version of HyperDEX that is somewhat tailored to MM2 and contains a script that automatically fetches or updates the latest MM2 binary from http://195.201.0.6/etomic?

You mean at runtime? I'm not a big fan of that idea. We can instead just have our nightly build use the latest build of mm v2 every night.

ArtemGr:
`> Let's not start discussing wild hypotheticals. If someone breaches the GitHub protected branch feature, there are many other projects much worse affected by that than us. What if someone tampers with the Rust compiler, or you know the CPU ;) There's always a risk, and I do agree, committing does not give much extra security benefit.`

We're discussing security, comparing different solutions. Making hypothesis, a thought experiment, is a valid approach towards doing such comparison. Do you have a different approach in mind?

`> Use a separate repo for it like we do? We would prefer a secure and trusted endpoint to fetch mm v2 builds.`

We'll be getting the binaries from http://195.201.0.6/etomic/, since this is where they are created. Uploading them to GitHub simply adds another link to the chain, without making it any more secure.

To make it more secure we can add a TLS cert to the server.

Again, what is the point of uploading them to GitHub? How is it more secure?

We can probably do this, of course, but I presently fail to see the point.

sindresorhus:
> GitHub has been hacked to this end before

Not to my knowledge. Source? The Rails Git repo was hacked once, but that was way before protected branches and 2FA.

> administrative privileges (GitHub has been hacked to this end before), disable the master branch protection, tamper with the Git history, then re-enable the protection.

We enforce 2FA in the org, so they would have to somehow bypass 2FA. It would also show up in the audit logs: https://github.com/organizations/atomiclabs/settings/audit-log?utf8=%E2%9C%93&q=action%3Aprotected_branch.destroy+

> but I don't think there's any proof that GitHub commits are more tamper-proof than GitHub release binaries uploaded otherwise.

And I agree with you on that.

ArtemGr:
I would strongly prefer **not** to have multiple nightly MM2 releases in circulation. When an early adopter comes to us on Discord or anywhere and tells us that he discovered a bug or a problem, I'd like to be able to work from the assumption that she has the latest or very recent MM2 release. The idea to bundle the MM2 releases into the HyperDEX releases is, I think, detrimental to the efficiency of the nightly MM2 testing and support.

artemii235:
Not sure about security concerns - ofc it's important, but there is always a flaw. As we talk about nightlies I think we SHOULD care about security but making sure that every tester have the latest build is more important. Sometimes we do dozens commits per day and HyperDEX nightly release can become outdated in few minutes after publishing.
Moreover if we enable full paranoia mode nightly builds (and releases too) are not secure already as these are built by CI services that we DON'T have control of and uploaded to service that we DON'T have control of. 
I think implementing a loader for MM2 at least for nightly builds is great idea, that's the only way to make sure testers always use fresh version.

ArtemGr:
And we can implement the public key verification of binaries in the loader.

ArtemGr:
NB: https://github.com/artemii235/SuperNET/issues/220 should now be fixed (that is, RPC errors will be reflected with the 500 HTTP status).
Etomic branch builds will likely appear tomorrow, when @artemii235 would take a look.

artemii235:
@ArtemGr Thanks! I'll merge it today.

sindresorhus:
I'm not against a binary loader. It would just be more work on our part. With nightly releases, we could reuse the existing Electron auto-updating functionality.

How would the binary loader work? Should it be implemented on our side or do you intend to make mmv2 self-updating? I assume on the first start HyperDEX would use the bundled mm2, then in the background fetch the latest one, but that means it will always be behind. A long-running daemon, like Chrome would be better, but it's complicated to do cross-platform, especially with Electron.

ArtemGr:
> I'm not against a binary loader.

Woot!

> It would just be more work on our part. With nightly releases, we could reuse the existing Electron auto-updating functionality.

We're generating the marketmaker-mainnet binaries already, generating a second binary that'll 1) http-fetch the MarketMaker archive (or skip it if its Last-Modified is old); 2) unpack; 3) cryptographically verify - is, IMHO, comparable in complexity with setting up the Electron auto-updating solution, while being more flexible and generic in that anyone can grab the loader and use it (with dICO, BarterDEX, HyperDEX, AtomicDEX or manually from command line).

> How would the binary loader work?

The first version will simply HTTP-fetch the nightly CI build, unpacking, verifying and transparently executing it. This will happen transparently, e.g. the GUIs and users will be getting the nightly loader and will be running it as they are running the MarketMaker today, passing it the configuration JSON on the command-line. The only change visible to the GUIs and users will be the download delay when the loader starts for the first time or updates.

> Should it be implemented on our side or do you intend to make mmv2 self-updating?

Planning to work on the nightly loader soon as we get the NAT traversal working for swaps.

> I assume on the first start HyperDEX would use the bundled mm2, then in the background fetch the latest one, but that means it will always be behind.

For the nightly releases it might be preferable to simply download the fresh version if there is one, e.g. updating in the foreground rather than in the background.

> A long-running daemon, like Chrome would be better, but it's complicated to do cross-platform, especially with Electron.

Background updates might make sense for a stable, production-grade MarketMaker.

We have some interesting options here, like using libtorrent ( cf. https://discordapp.com/channels/412898016371015680/476025090627207168/513750386822217730 ) to
a) Securely distribute the updates (through the DHT) without any central servers or points of failure. That is, while we still own the private keys, we can distribute the updates from anywhere, even if the GitHub or CI infrastructure moves or fails.
b) Use Bittorent peer swarms to speed-up the downloads and reduce the load on the seed servers.
c) Download the updates at reduced speed, in order to interfere less with anything that user's doing in the foreground.
This is maybe how it _should_ work when the atomic swaps are vastly popular and everyone's grabbing the MM. Less need to implement something like this for now. But binary loader leaves this path easily open to us.

Closer to where we are, I think the immediate goal is to
1) Have a simple foreground loader just for the MM2 _nightlies_, that you would deploy with a special HyperDEX version and that would work transparently like a usual MM2 binary.
2) When we're ready to roll a stable MM2 branch, and _if_ we're pleased with how the loader works for the nightlies, _then_ consider using it for the stable branch too.

lukechilds:
I'm not sure auto-updating the binaries is a good idea. There are security concerns with auto-updating, especially when dealing with sensitive data like private keys.  Even if we implement signing/checksums to protect against MITM attacks I still don't like the idea of code I haven't manually approved being downloaded and installed on the machine that holds my private keys. An auto-updater is essentially a remote code execution vulnerability that's only open to a trusted entity.

One option (which we discussed previously https://github.com/atomiclabs/hyperdex/issues/9) would be opt-in auto updates, where the user has to manually confirm they consent to an update. That gives them the option to audit changes in the source code first.

There are also stability issues with HyperDEX fetching the latest mm builds. If you need to make an API change to fix something and we update HyperDEX to work with the new API, old HyperDEX builds are going to download and run new mm builds they can't properly communicate with. That sounds like it could cause a lot of headaches and false positives with testers.

I feel like we're over complicating this a bit. You just need a simple way for users to test mm via HyperDEX, right? And I believe you already have Docker images available for mmv2, is that correct? If so, what about if testers install HyperDEX and mmv2 seperately? They install a HyperDEX binary and mmv2 via Docker. Then configure HyperDEX to connect to the Docker instance of mm instead of the bundled mm binary.

That would give you full control over what version of HyperDEX and mm your testers are running. Would that be a good solution for you?

ArtemGr:
> I'm not sure auto-updating the binaries is a good idea. There are security concerns with auto-updating, especially when dealing with sensitive data like private keys. Even if we implement signing/checksums to protect against MITM attacks

We will.

> I still don't like the idea of code I haven't manually approved being downloaded and installed on the machine that holds my private keys.

Yeah, I get it, but I'd argue that it's more of a perception than a security issue.

To test the MM2 you have to fetch it from http://195.201.0.6/etomic/. By manually fetching it from there, you're getting this nice feeling that you've manually approved the download. But it doesn't have any security value, if, for example, the CI servers are compromised and these binaries have backdoors.

Similarly, distributing your application through GitHub you're getting this warm fuzzy feeling knowing that you're doing what pretty much everyone else on this planet does. But from the security standpoint all it does it gives the high-level attackers another vector to tamper with the binaries.

> An auto-updater is essentially a remote code execution vulnerability that's only open to a trusted entity.

You are exactly right! And in the same vein, any binary download is essentially a remote code execution vulnerability. The fix is to move to operating systems that would routinely sandbox the binaries. We're not quite there yet.

> One option (which we discussed previously #9) would be opt-in auto updates, where the user has to manually confirm they consent to an update. That gives them the option to audit changes in the source code first.

I don't want to see outdated versions of MM2 in circulation.

So yeah, for stable builds I guess we can make the auto-updates optional: we'll simply tell the user that her version of MM2 is outdated and that she needs to grab a new one.

> There are also stability issues with HyperDEX fetching the latest mm builds. If you need to make an API change to fix something and we update HyperDEX to work with the new API, old HyperDEX builds are going to download and run new mm builds they can't properly communicate with. That sounds like it could cause a lot of headaches and false positives with testers.

This shouldn't happen regardless of the binary distribution (and is a separate concern), cf. https://semver.org/.

> I feel like we're over complicating this a bit. You just need a simple way for users to test mm via HyperDEX, right?

We need users to have a way of test a **fresh** version of MM2 via the HyperDEX, yes.

> And I believe you already have Docker images available for mmv2, is that correct? If so, what about if testers install HyperDEX and mmv2 seperately? They install a HyperDEX binary and mmv2 via Docker. Then configure HyperDEX to connect to the Docker instance of mm instead of the bundled mm binary.

I think this over-complicates things a bit. Users can simply grab the fresh MM2 binary from http://195.201.0.6/etomic/ and replace the existing one with the new one. No need for Docker and all.

But I'm of an opinion that having a special HyperDEX release that one can simply grab and install as usual, without tampering with MM2 binaries, is vastly superior in user experience.

> That would give you full control over what version of HyperDEX and mm your testers are running.

It won't, Docker isn't self-updating.

> I would suggest the Docker + HyperDEX is still the simplest and cleanest solution, however if you want to implement auto updating directly inside mm then I guess that's up to you.

Thanks for the idea, Luke!
Would be nice to have a version that needs no tampering though.

> And I suppose the security concerns aren't really valid for testing, as long as you notify the testers and make them aware.

The GitHub distribution is strictly less secure than the direct downloads from http://195.201.0.6/etomic/.

ArtemGr:
> Early on in development I often needed to run different branches of mm to trouble shoot stuff with James so we added the option to connect to an external mm url.

The goal here is to give something for early adopters to install and run as usual, without external MMs and their URLs.

lukechilds:
Yeah, I was just recommending that for you to use for development, then we will do a new build of HyperDEX with the auto updating binary included:

>Maybe it would be best for you to develop auto-updating and test using this method to get it working. Once that's done we can do a new build of HyperDEX for MM2 testing that bundles auto-updating MM2.

ArtemGr:
Got it, thanks!

artemii235:
@sindresorhus @lukechilds Please set the default netid for MM2 to 9999. We need MM2 to always be on non-production netid as it's not compatible with existing MM1 nodes already.

sindresorhus:
@artemii235 Done

sindresorhus:
I tried updating to the latest mmv2 build (dddff9f / macOS) and I now get this error when trying to start Marketmaker through HyperDEX:

```
Error: Could not get ETOMIC from faucet!
```

```
      mm › 10 20:49:16, etomiccurl:287] Etomic faucet request: "{\"etomicAddress\":\"RCWSXK8DXavzp29BZErXqpADpr5S1Y8Yke\"}"
      mm › 10 20:49:17, etomiccurl:295] Got error from Etomic faucet: "Error occurred"
      mm › 10 20:49:33, etomiccurl:287] Etomic faucet request: "{\"etomicAddress\":\"RCWSXK8DXavzp29BZErXqpADpr5S1Y8Yke\"}"
      mm › 10 20:49:33, etomiccurl:295] Got error from Etomic faucet: "Error occurred"
```

artemii235:
@sindresorhus Hi, the faucet was drained again, refilling now.

artemii235:
@sindresorhus Faucet is refilled, please recheck.

sindresorhus:
@artemii235 Thanks, works fine now.

sindresorhus:
Continued in https://github.com/atomiclabs/hyperdex/pull/602.

PR #565, opened 2018-11-01
Add more logging

sindresorhus:
And refactor the `request` method.

This will give us a nice overview of what HyperDEX is doing regarding communication with Marketmaker and even how long it takes. I initially needed this for #563, but it's useful for mm v1 too.

By combining `_request` and `request`, I managed simplify it. I also had to do this to get access to the data I needed at the same time for logging.

It shows both the request and the response:

screen shot 2018-11-01 at 23 57 45


sindresorhus:
Merging now as I need it for mm v2 builds, but still open to feedback.

PR #566, opened 2018-11-02
Stop filtering order books based on UTXO count

sindresorhus:
Since it's going away in mm v2 and it's not that useful or accurate anyway.

I'm trying to land as many changes on master as possible so we don't diverge too much, since I expect the mm v2 branch to be long-running.

### DIFF
--- a/app/renderer/api.js
+++ b/app/renderer/api.js
@@ -173,12 +173,10 @@ export default class Api {
 		});
 
 		const formatOrders = orders => orders
-			.filter(order => order.numutxos > 0)
 			.map(order => ({
 				address: order.address,
 				depth: order.depth,
 				price: order.price,
-				utxoCount: order.numutxos,
 				averageVolume: order.avevolume,
 				maxVolume: order.maxvolume,
 				zCredits: order.zcredits,


lukechilds:
Filtering out `order.numutxos > 0` is quite useful for now, if a Bob runs out of UTXOs he can't trade, but his orders will still show in the order book, just with `order.numutxos == 0`.

Without this filter we will list orders in the order book which can never be filled. This would be confusing for the user because they'll see orders that they'll try and price match but a match will never occur.

It also means you could do a "kind of" DoS attack by spinning up a Bob with 0 UTXOs. Because it's the lowest price, most users would match this price and it'll never go through. It also means clicking the target button which auto-selects the best price will never match either.

I don't think we should remove the feature that prevents this from the current builds just because mmv2 may not need it. If mmv2 is removing `order.numutxos` then it also needs to implement something else to notify peers of orders that can't be filled. I think we should wait to see what that looks like before making code changes to accommodate it.

To clarify, it's fine to make this change in the mmv2 branch to accommodate for the mmv2 behaviour during testing. But I don't think it makes sense to merge it into the main release.

sindresorhus:
Good points. I agree. I didn't think of this. I'll merge it in the mm v2 branch.

Issue #454, opened 2018-07-27
Export/import portfolio


sindresorhus:
It could be useful to be able to export a portfolio (encrypted seedphrase, enabled currencies, trade history etc) and import it at some other computer or just for backup reasons.

We can probably just put all the info in a JSON file.

lukechilds:
You could do it now by just zipping/unzipping the app data folder.

But obviously that's a bit heavy and possible error prone.

Swap history + portfolios + config smushed into a single import/export JSON file is much nicer.

PR #575, opened 2018-12-04
Move enabled currencies from the app config to the portfolio

### LINKED ISSUE #471: Store the enabled currencies in the portfolio itself, not in the app settings file
This will be important for when we allow import/export of portfolios (#454), but also now, so new portfolios get new default currencies we add (https://github.com/atomiclabs/hyperdex/issues/469#issuecomment-412145712).

sindresorhus:
Fixes #471 and paves the way for #454.

Includes automatic migration.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -1,11 +1,7 @@
 'use strict';
-const {api, is} = require('electron-util');
+const {is} = require('electron-util');
 const Store = require('electron-store');
 const {minWindowSize} = require('./constants');
-
-// Note: This is hardcoded here instead of being imported
-// from `util-common` to prevent cyclic dependencies
-const isNightlyBuild = api.app.getName() === 'HyperDEX Nightly';
 
 const defaults = {
 	windowState: {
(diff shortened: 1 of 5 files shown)


kevva:
The first time I logged in I got the error below. After trying again it worked fine. Adding/removing currencies also works and shows up in the portfolio file as expected.

screenshot 2018-12-05 at 18 11 46

sindresorhus:
@kevva Thanks, missed that. Fixed.

kevva:
Hmm, I still get the error. To reproduce; remove `currencies` from the portfolio file and try logging in.

kevva:
I get this when starting the app now:

screenshot 2018-12-08 at 15 54 05

sindresorhus:
@kevva I'm still unable to reproduce your issue, but I've tried a different approach to fix https://github.com/atomiclabs/hyperdex/pull/575#issuecomment-444566947. Can you reproduce the missing portfolio issue with the latest commit?

kevva:
@sindresorhus, strange. I got it sometimes, and sometimes it worked. However, with this new solution, it never seems to occur so I think this should be ready now! Great work!

PR #577, opened 2018-12-07
Update Electrum servers for some currencies and remove some currencies

sindresorhus:
- Remove ARG: https://github.com/jl777/coins/commit/cef0e82b62ac95599bfdd92cde0432dbcddbc65d
- Remove MAC: https://github.com/jl777/coins/commit/8e49e8667114659f30132a36ebd305ee1c8af748
- Remove CALL: https://github.com/jl777/coins/commit/77c6c0a83c293bc15f0d8de4567956567e781303
- Remove DNR: https://github.com/jl777/coins/commit/791e3be03cef93c410bcb6c4a39b1f88ebc54d5f
- Update LTC Electrums: https://github.com/jl777/coins/commit/9316e5a6792a0aad1bd1127968127b5ff9ccd72b
- Update BCH Electrums: https://github.com/jl777/coins/commit/89687197fae59bb4383c2b7337b1f41472693527
- Update ZEC currency data: https://github.com/jl777/coins/commit/178caa1ec5850ca2d2b854cfb584f1e92f2f4365


sindresorhus:
For `ARG`, `MAC`, and `DNR`, I assume it was just removed because @cipig doesn't want to host their Electrum servers anymore?

sindresorhus:
- [ ] Just noticed that DNR is in our list of default currencies, so I need to create a migration step for that if we decide to remove it.

cipig:
- had to stop ARG because i can't compile their wallet any more. Wallet had difficulties in the past. Wallet and electrum server consumes much RAM.
- MAC wallet consumes up to 6GB RAM, electrum servers are barely used, better use this ressources for other coins
- DNR electrums were run by the coin owner, but not maintained, so they were causing problems
- i think we should not waste too much ressources for external coins

lukechilds:
@cipig are the servers already down or do we have some time to warn users?

As far as they're concerned we're effectively delisting the currency and locking their funds up as soon as it's delisted.

Are you aware of any community servers we can point to instead?

cipig:
ARG and MAC are down and there are no running public servers
LTC and BCH are running, but i will stop them before updating KMD wallets (running out of RAM)
LTC and BCH have enough public servers
i don't think it should be our responsibility to run electrums for each and every coin

Issue #591, opened 2019-02-07
Coin servers


jdmarlow86:
**Describe the bug**

GLXT and MNZ servers are down

**To reproduce**
1. Log into HyperDEX

**Expected behavior**

Login is successful with no errors

**Screenshots**

![image](https://user-images.githubusercontent.com/22251770/52389855-8a2f5f00-2a63-11e9-8760-bf7163a84c26.png)

**System info**
- OS & version: WIN10
- HyperDEX version: 0.4.0

lukechilds:
This is an issue with the MNZ and GLXT electrum servers, not HyperDEX.

We don't run the Electrum servers. Does HyperDEX still log in ok?

sindresorhus:
@cipig removed the MNZ Electrum servers (https://github.com/jl777/coins/commit/89b873b285493d79e026c0dee215227a522f3cd7), so I guess they're offline. Unclear why they were removed though.

cipig:
MNZ is dead. Asked ca333 and he told me i can remove the electrums, which i did.

jorian:
GLXT and MNZ are dead and removed.

PR #592, opened 2019-02-19
Remove Monaize (MNZ) currency

sindresorhus:
It's dead, see #591.


lukechilds:
Thanks @sindresorhus!

PR #593, opened 2019-02-19
Upgrade `ow` dependency

sindresorhus:


lukechilds:
That label syntax is much nicer 👌

sindresorhus:
And when used in Node.js, you don't even need the label as it's automatically inferred 👌

PR #595, opened 2019-03-12
Added new cmcticker coin attribute - can be used to override the defa…

FreeTrade:
…ult ticker when requesting quotes from CoinMarketCap. See HOdlcoin for an example.

Re-enabled HOdlcoin with the HODLC/HODL ticker

### DIFF
--- a/app/marketmaker/supported-currencies.js
+++ b/app/marketmaker/supported-currencies.js
@@ -1066,7 +1066,6 @@ const supportedCurrencies = [
 			},
 		],
 	},
-	
 	{
 		coin: 'HUSH',
 		rpcport: 8822,
--- a/app/marketmaker/supported-currencies.js
+++ b/app/marketmaker/supported-currencies.js
@@ -1026,7 +1026,9 @@ const supportedCurrencies = [
 	/* Disabled because of #289
 	{
 		coin: 'HODL',
+		name: 'HODL (Komodo Asset)',
 		asset: 'HODL',
+		cmcticker: 'HODLKA',
 		rpcport: 14431,
 		electrumServers: [
 			{
@@ -1043,9 +1045,11 @@ const supportedCurrencies = [
 			},
 		],
 	},
+	*/
 	{
 		coin: 'HODLC',
-		name: 'Hodlcoin',
+		cmcticker: 'HODL',
+		name: 'HOdlcoin',
 		rpcport: 11989,
 		pubtype: 40,
 		p2shtype: 5,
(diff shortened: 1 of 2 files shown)


app/marketmaker/supported-currencies.js
```diff
@@ -1026,7 +1026,9 @@ const supportedCurrencies = [
 	/* Disabled because of #289
 	{
 		coin: 'HODL',
+		name: 'HODL (Komodo Asset)',
 		asset: 'HODL',
+		cmcticker: 'HODLKA',
```
sindresorhus: This is not a ticker on CMC.

app/marketmaker/supported-currencies.js
```diff
@@ -1043,9 +1045,11 @@ const supportedCurrencies = [
 			},
 		],
 	},
+	*/
 	{
 		coin: 'HODLC',
-		name: 'Hodlcoin',
+		cmcticker: 'HODL',
```
sindresorhus: We don't use CMC anymore, so we need a more generic name for this property.

sindresorhus (changes requested):

app/marketmaker/supported-currencies.js
```diff
@@ -2268,6 +2271,11 @@ const getCurrencyName = symbol => {
 	return coinParams.name || coinlist.get(symbol, 'name') || symbol;
 };
 
+const getCurrencyCMCTicker = symbol => {
```
sindresorhus: Same as https://github.com/atomiclabs/hyperdex/pull/595/files#r286099824

app/marketmaker/supported-currencies.js
```diff
@@ -2268,6 +2271,11 @@ const getCurrencyName = symbol => {
...
+	const coinParams = supportedCurrencies.find(currency => currency.coin === symbol);
```
sindresorhus: `coinParams` => `currency`

app/marketmaker/supported-currencies.js
FreeTrade: Yes, asset HODL does not have a CMC ticker, but one must be specified so that it does not automatically use the HODL symbol and gather incorrect data if this asset is re-enable in the future.

app/marketmaker/supported-currencies.js
FreeTrade: symbols/tickers are different at different exchanges, otherwise we'd just use the 'coin' property in all cases. What are we using in place of CMC?

PR #600, opened 2019-05-11
Use `PropType.elementType` where possible

### LINKED ISSUE #598: Use the new `elementType` PropType
See https://github.com/facebook/prop-types/pull/211

We should use it instead of `PropType.func` in most places, except for change handlers.

ru-lai:
…e the elementType PropType instead of PropTypes.func
@sindresorhus 
@lukechilds 

Fixes #598


sindresorhus:
This looks good. Thank you for contributing, @ru-lai 🙌

PR #614, opened 2019-07-30
Add ability to search currency selectors in Exchange view

### LINKED ISSUE #607: Exchange view currency filter
When in exchange view, you need to select the coins from the list on either side. Would be good to allow a user to start typing the name of the coin and it will filter based on that.

sindresorhus:
Screen Shot 2019-07-31 at 01 25 43

Fixes #607

### DIFF
--- a/app/renderer/components/Select.scss
+++ b/app/renderer/components/Select.scss
@@ -277,8 +277,11 @@
 	border-bottom: none;
 }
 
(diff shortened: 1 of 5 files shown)


lukechilds:
Lovely stuff.

Build is failing due to lint rules but the failures look unrelated to this PR.

